PR #61, opened 2016-09-12
Added new Pat installer package template, and related resources.

kd8drx:
Updated the Pat Packages template to have nicer UX, to display a license agreement, and to generate a signed package if possible.


martinhpedersen:
Thank you Will!

I'll do some minor modifications if you don't mind:
1. Reference `../pat` instead of `/usr/local/bin/pat`
2. Use `../LICENSE` instead of `Pat-License.txt`
3. Remove the `<key>CERTIFICATE</key>`-section so that others are able to build without having the correct certificate.

Number 3 does unfortunately mean that you'll have to manually sign the package, but in return I've added a `packagesbuild` step to make.bash so that you only need to do `./make.bash && productsign ...`. Hope that's ok!

kd8drx:
Martin,
#3 is totally fine - that’s something I should have done myself.

re #2: That was my first thought as well, however, the line breaks in LICENSE cause some odd formatting in the Installer package. Not huge, but annoying enough that I made a dedicated copy to address. An alternative would be to remove the line breaks in LICENSE.

for #1, are you referring to the info in Pat-info.rtfd, or where Packages is sourcing the binary?

^WPD

martinhpedersen:
> That was my first thought as well, however, the line breaks in LICENSE cause some odd formatting in the Installer package.

Ah, I see. Removing all the line breaks in LICENSE does not look so good when reading it in the terminal. I will leave this as you intended :)

> for #1, are you referring to the info in Pat-info.rtfd, or where Packages is sourcing the binary?

The latter - the location where Packages is looking for the binary. I will verify that I'm not changing the install path before committing the change. Are you ok with this?

kd8drx:
That should be fine - my concern was what you stated: we want to make sure the binary ends up in /usr/local/bin with the right permissions.

martinhpedersen:
Excellent! And again, thank you for this.

PR #67, opened 2016-10-12
Release/v0.2.1

martinhpedersen:


martinhpedersen:
The linux binaries will no longer require libax25 as a shared library - it is statically linked. This should be reflected in the wiki.

PR #76, opened 2017-03-16
view winlink forms (WIP)

Docteh:
I did some work towards winlink form support, and I figured I'd put it up as a pull request.

The main thing missing, is a configuration option for where to look for the form files, and I think the interface should be modified to have a view option along side the download. 

disclaimer: I didn't run "go fmt"


martinhpedersen:
Kyle - thank you for this pull request. Very exciting!

I haven't got a chance to review it yet, but it sounds like a very nice feature to include:) I'll look into it as soon as possible.

Would you be willing to add the missing config option and the UI view option if I point you in the right direction? 

Every PL is greatly appreciated, and I have no problem implementing the last parts myself if you need me to:)

Docteh:
Hi Martin, I'll give it my best shot, if it turns out really bad we can always turf it haha

For the bindata, I found https://github.com/jteeuwen/go-bindata but it took me awhile to read their instructions for including files recursively 

go-bindata -o bindata_assetfs.go res/...

martinhpedersen:
Kyle - sorry for the delayed response. I am very excited about this, but unfortunately the last couple of weeks have been very busy.

Had some minutes to check out your commits tonight, but I am having problems opening the one form XML attachment I have in my inbox:
```
2017/03/23 21:28:53 html/template:t: expected space, attr name, or end of tag, but got "=\"1\" style=\"font-weight: normal;\"><br />&nbsp; &nbsp; For&nbsp; &nbsp; &nbsp;</pbzloc></span><input style=\"color: rgb(204, 0, 0);\" readonly=\"readonly\" value=\""
```

Do you have any idea what went wrong? The XML data was generated by Winlink Express a couple of weeks ago, and the form viewer should be up to date as I downloaded the latest copy from winlink.org today.

Maybe you could send me an example message I could use to test with? LA5NTA@w..... :)

Also, we probably need to handle these kinds of errors (malformed/incompatible XML form data) more gracefully, maybe something like a 5xx status code and some human readable error message explaining the cause of the error (malformed XML data, incompatible form viewer etc). This particular error caused Pat to exit (log.Fatal does that), which seems a bit harsh ;)

Docteh:
Hi Martin, 
Simple_Bulletin_Viewer.html

Looks like the templating system is choking on the html file. <pbzloc ="1"... In one tag they've specified a value, but no attribute name. 

I'm actually confused, like this is clearly an HTML formatting issue, I just was not expecting the templating engine to notice. 

I noticed that template.execute mentions that it may exit after some output has been sent, so I'll probably want to buffer the output. 

I'm pushing a small change to fix error reporting while I think about this.

-Kyle

Docteh:
im going to do some research on this pbzloc element, see if I can understand it a bit better. From what I can tell it's not a real tag.

Right now in my code I'm rewriting the templates so I can use the built in template package, but I could add a bit more code and skip the template engine.

martinhpedersen:
Hi Kyle - did your receive my winlink message response?

The forms I've tested works perfectly now. My hope is to include this in the next release, but I think we need to work some more on the GUI's view button and probably do some more testing and error handling.

Regarding the GUI. Instead of having two buttons, I am thinking a [split button with dropdowns](https://v4-alpha.getbootstrap.com/components/dropdowns/#split-button-dropdowns) providing attachment actions (view, forward, download etc.). What do you think about that?

> Right now in my code I'm rewriting the templates so I can use the built in template package, but I could add a bit more code and skip the template engine.

What would be the gain of skipping the template engine?

--------

In time, we may benefit from refactoring the winlinkviewHandler into a function that is better suited for CLI integration. Maybe we could move the logic to an internal package like [internal/gpsd](https://github.com/la5nta/pat/tree/master/internal/gpsd) and write some unit tests.

Docteh:
Hi Martin, sorry I've been distracted with some sunny weather. 

I originally used the template engine, because the documentation for it mentioned that when inserting variables it will escape the right characters depending on where it is in the html document. After looking into Winlink forms a bit more in depth, such complexity isn't needed. Web browsers are pretty forgiving when it comes to dealing with bad HTML. 

Two things I need to fix, one minor thing i noticed winlink express doing, if a variable doesn't exist, it doesn't get replaced eg put {var ImBatman} in a view. This could be ignored. The other is newlines in the XML need to be replaced with <br> tags.

martinhpedersen:
Kyle - Hehe, sunny weather here in Norway too these days, so no need to apologize to me :wink: 

OK, I understand. Then I would prefer that we avoid using the template engine if you don't mind.

I plan to have a go at the split-button approach for the view button soon. It would be great if you could include the fixes you mention, but no rush. :+1:

martinhpedersen:
@Docteh - are you still interested in working on this feature?

http.go
```diff
@@ -41,6 +44,7 @@ func ListenAndServe(addr string) error {
 	r.HandleFunc("/api/mailbox/{box}", mailboxHandler).Methods("GET")
 	r.HandleFunc("/api/mailbox/{box}/{mid}", messageHandler).Methods("GET")
 	r.HandleFunc("/api/mailbox/{box}/{mid}/{attachment}", attachmentHandler).Methods("GET")
+	r.HandleFunc("/api/mailbox/{box}/{mid}/{attachment}/view", attachmentViewHandler).Methods("GET")
```
programmin1, line 47: Is this only related to _viewing_ a sent Winlink form not filling a winlink form based on the forms users can pull in from https://www.winlink.org/content/how_manually_update_standard_templates_version_101200 ? 
I had some ideas on sending as it looks like that is somewhat web based now, don't want to reinvent.

Docteh: Hi, this code is just for viewing forms.

programmin1: That's what I thought. As soon as I can [get this to build locally](https://github.com/la5nta/pat/issues/195) I have some ideas on adding filling out the forms and sending them, feature. If you haven't added that already?

martinhpedersen:
FYI
See this issue: https://github.com/la5nta/pat/issues/135#issuecomment-596306591
And this dev branch: https://github.com/rainerg2000/pat/tree/issue/135-support-ics-forms

Thanks!

PR #170, opened 2019-10-07
Preview position report with OpenLayers+OSM

martinhpedersen:
Use OpenLayers and OSM tiles to display a preview of the current position in the Position Report modal.

Currently the user would need to be online to fetch tiles, but I believe it would be possible to extend this to use a custom (local) tile server. I'm hoping browser cache will help in some cases as well.

**Anyone interested in such a feature?**

(ping @blockmurder)

-----------

Skjermbilde 2019-10-07 kl  20 01 44


blockmurder:
Hi Martin,
seems like quite a nice feature! Is it possible to drag the pin to a new location? We sometimes are quite lazy or forget to post a position when we are sailing. So dragging the pin to an "old" location would actually be really nice too. One would have to be able to adjust the time/date too, though.
Also having the map offline is kind of a must whilst beeing offshore.

Would it be possible to have it configurable (e.g. enable/disable)?

I'll try it out tomorrow during the day!

res/js/index.js
```diff
@@ -346,12 +348,51 @@ function handleGeolocationError(error) {
...
+	return zoom;
+}
+
 function updatePositionGeolocation(pos) {
 	var d = new Date(pos.timestamp);
 	statusPos.html("Last position update " + dateFormat(d) + "...");
 	$('#pos_lat').val(pos.coords.latitude);
 	$('#pos_long').val(pos.coords.longitude);
 	$('#pos_ts').val(pos.timestamp);
+	updatePosMarker(null);
```
blockmurder: Instead of calling `updatePosMarker(null)` after each position update, you could consider the ajax complete callback :
```
complete: function( jqXHR, textStatus ){
	updatePosMarker(null);
}
```
https://github.com/la5nta/pat/blob/7bf89d2beb5700111bcff3c5227dc47af0d6dfa4/res/js/index.js#L54-L77

Otherwise the map div does not show the map at all (if no manual imput, GPSd and Geolocation unavailable).

res/js/index.js
```diff
@@ -346,12 +348,51 @@ function handleGeolocationError(error) {
 	statusPos.html("Geolocation unavailable.");
 }
 
+var map = null;
+var markers = null;
```
blockmurder: Maybe add default marker (same coordinates as placeholder: 60.1234, 5.0123) to be updated if manual input/GPSd/Geolocation available.

martinhpedersen:
I'm closing this one. It does seem like a nice feature, but I feel like we need a solution for offline maps first(?). I think it's best if I prioritize other features for now.

Feel free to continue this work if anyone is interested :)

Thanks for your feedback @blockmurder 👍

Issue #175, opened 2019-12-07
Add wiki page on how to hack on the source code


martinhpedersen:
Ref https://github.com/la5nta/pat/pull/173. We should add some wiki page to help new contributors how to hack on the source code.

Off the top of my head:
* Dependency management (using git submodules)
* How to maintain/update `bindata_assetfs.go` and the web sources
* How to add new transports
...

xylo04:
I added a note about `go generate` to https://github.com/la5nta/pat/wiki/Building-from-source

KI7ODK:
Yes, having some info on adding new transports would help me with adding AGW.

martinhpedersen:
Tyler, I will add some info as soon as possible for you. I'll let you know when it's done.

martinhpedersen:
https://github.com/la5nta/pat/wiki/Adding-transports

Hopefully enough to get you started :) Please let me know if you have further questions. I guess we can continue the discussion at https://github.com/la5nta/wl2k-go/issues/57, and hopefully we can remember to update the wiki as we go.

martinhpedersen:
I think we have enough documentation covering this topic now:
* https://github.com/la5nta/pat/wiki/Building-from-source
* https://github.com/la5nta/pat/wiki/Adding-transports

Closing this 🎉

PR #178, opened 2019-12-17
Add 'go generate' to make.bash

xylo04:
This will make web development easier by making resource regeneration part of the regular build process. There's also a 'go-bindata-assetfs -debug' flag I'd like to make supported via environment variables, but I can't figure out how to do it cleanly yet, so I'll save that for later.

Also add docs about not submitting changes to bindata_assetfs.go.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,7 +9,7 @@ environment:
     GOVERSION: "1.13.1"
     MSYS_PATH: C:\MinGW\msys\1.0
 install:
-  - set PATH=C:\go\bin;%MSYS_PATH%\bin;C:\MinGW\bin;%PATH%
+  - set PATH=%GOPATH%\bin;C:\go\bin;%MSYS_PATH%\bin;C:\MinGW\bin;%PATH%
   - rmdir c:\go /s /q
   - appveyor DownloadFile https://dl.google.com/go/go%GOVERSION%.windows-386.zip
   - 7z x go%GOVERSION%.windows-386.zip -y -oC:\ > NUL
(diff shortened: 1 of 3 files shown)


xylo04:
It seems like AppVeyor isn't set up to include `GOPATH/bin` in `PATH`, is that something that can be configured?

xylo04:
Related to #175; having this part of the automatic build process reduces the need for prose documentation, although reviewers may need to remind contributors to revert changes to `bindata-assetfs.go`.

xylo04:
Thanks for the review. I understand your hesitance.

> It changes the «source code» upon compilation,

As a generated artifact, I'd consider `bindata_assetfs.go` as only "pseudo" source code.

> making reproducible builds less likely.

If you really want to move toward true [reproducible builds](https://reproducible-builds.org/), one of the prerequisites is that timestamps cannot be encoded in artifacts. Between `go-bindata-assetfs`'s date behavior and git's lack of date preservation, that's basically impossible for us. Without overhauling that part of the toolchain, we have to settle for mostly reproducible.

> It introduces a new dependency to the build chain which most users won’t really need to compile the program.

I'd argue it's a lightweight build dependency in that it installs itself via `go generate`, but point taken.

> What about using build tags to switch between loading resources from disk or bindata_assetfs.go?

I hadn't seen build tags before. They look like an interesting mechanism, but I'm not sure how it would interact with `go-bindata-assetfs`. The `-debug` flag I described replaces `bindata_assetfs.go` with shim methods that read from the local disk, and it's done by absolute paths, so each developer needs to do it for their own system. It's not something we could check in alongside `bindata_assetfs.go` with build tags.

----

I'm not adamant that this particular change or any change be made to automate regenerating web resources, it was just a stumbling block that I experienced and I wanted to prevent that for other new devs. It's now documented in the wiki. Without switching to a fully-featured build system like Bazel, I think we'll be faced with compromises like this.

I'm content to close this PR unless you want to do something else.

Issue #156, opened 2019-05-06
Web GUI - make options more intuitive


DC7IA:
Some of the options in the web GUI are not really intuitive.

As an example: The checkbox "Radio Only". (Can someone tell me what it does?)
The man page tells me: `--radio-only         Radio Only mode (Winlink Hybrid RMS only).`
I thought I need a radio anyway to do amateur radio? What does that mean?

Also the dropdown menu "Select Alias". What is that good for? I only see telnet there..?

Compose -> new message: checkbox "P2P Only" - Maybe write there what it means? Does it mean that the message can't go via an automatic station?

I think pat will be far more user friendly with some descriptions for these options. :)

martinhpedersen:
Regarding Radio Only, a quick google search for "Winlink Hybrid RMS" gave me this decent explanation: http://www.philsherrod.com/Winlink/Winlink_Operational_Modes.pdf.
I'm not sure what more you need? Do you wan't the man page to give detailed information on how the Winlink system operates?

I agree that some of the options could be better documented and maybe need improved labels and/or descriptions.

Specific suggestions for improvements are much appreciated!

davistroy:
DC7IA, take a look at KM4ACK’s videos on YouTube.  He does an awesome job explaining all of it.

guitarpicva:
Martin, are you able to use tool tips in the web UI?  Those are helpful for short explanations.  Some people hate them, but I have trained my end users to look for them now.  Saves a lot of documentation for simple items.

martinhpedersen:
Tooltips are definitely an option :) It would be nice to be able to provide some more context to some of the buttons and menus.

I think tooltip support is limited with our current Bootstrap 3.2, but it seems possible using a plugin: https://getbootstrap.com/docs/3.4/javascript/#tooltips

Bootstrap 4 have better support for tooltips, and PR #197 will bring us up to date there I believe.

PR #180, opened 2020-01-10
Remove the "Actions" dropdown menu

bitmux:
A flatter menu seems more intuitive to me and was requested in bug #156

### DIFF
--- a/res/tmpl/index.html
+++ b/res/tmpl/index.html
@@ -35,19 +35,15 @@
             <li><a href="#" id="outbox_tab">Outbox</a></li>
             <li><a href="#" id="sent_tab">Sent</a></li>
             <li><a href="#" id="archive_tab">Archive</a></li>
-            <li class="dropdown">
+<!--          Flattened menu as the additional dropdown was annoying 
+              <li class="dropdown">       
               <a href="#" class="dropdown-toggle" data-toggle="dropdown">Action <span class="caret"></span></a>
               <ul class="dropdown-menu" role="menu">
-                <li><a href="#" data-toggle="modal" data-target="#connectModal"><span class="glyphicon glyphicon-refresh" /> Connect...</a></li>
-				<li class="divider"></li>
-                <li><a href="#" data-toggle="modal" data-target="#composer"><span class="glyphicon glyphicon-edit" /> Compose...</a></li>
-                <li><a href="#" data-toggle="modal" data-target="#posModal"><span class="glyphicon glyphicon-map-marker" /> Position...</a></li>
-                <li class="divider"></li>
-                <li class="dropdown-header">Other stuff</li>
-                <!--<li><a href="#">Settings</a></li>-->
-                <li><a href="#" data-toggle="modal" data-target="#aboutModal">About {{.AppName}}</a></li>
-              </ul>
-            </li>
+-->
+            <li><a href="#" data-toggle="modal" data-target="#connectModal"><span class="glyphicon glyphicon-refresh" /> Connect...</a></li>
+            <li><a href="#" data-toggle="modal" data-target="#composer"><span class="glyphicon glyphicon-edit" /> Compose...</a></li>
+            <li><a href="#" data-toggle="modal" data-target="#posModal"><span class="glyphicon glyphicon-map-marker" /> Position...</a></li>
+	    <li><a href="#" data-toggle="modal" data-target="#aboutModal">About {{.AppName}}</a></li>
           </ul>
         </div><!--/.nav-collapse -->
         <div id="navbar_progress" aria-hidden="true" style="display: none;">


martinhpedersen:
Ping @bitmux

DC7IA:
Just wanted to compile it and make some screenshots, but don't really know how to compile go source code. :/

```
go get -u github.com/bitmux/pat@patch-1

go: finding github.com/bitmux/pat patch-1
go: finding github.com/bitmux patch-1
go: finding github.com patch-1
go: finding github.com/gorhill/cronexpr latest
go: finding github.com/howeyc/gopass latest
go: finding github.com/nfnt/resize latest
go: finding github.com/pd0mz/go-maidenhead latest
go: finding github.com/paulrosania/go-charset latest
go: finding github.com/tarm/serial latest
go: finding github.com/tarm/goserial latest
go: finding golang.org/x/crypto latest
go: finding golang.org/x/sys latest
go: finding golang.org/x/net latest
../pkg/mod/github.com/bitmux/pat@v0.8.1-0.20200110033324-5f5ba52a483d/rmslist.go:22:2: use of internal package github.com/la5nta/pat/internal/cmsapi not allowed
../pkg/mod/github.com/bitmux/pat@v0.8.1-0.20200110033324-5f5ba52a483d/http.go:31:2: use of internal package github.com/la5nta/pat/internal/gpsd not allowed
```

The links are 404..? I wonder..?

martinhpedersen:
Thanks for helping out Joshua 👍 

You need to clone the repo to $GOPATH/src/github/la5nta/pat to resolve those errors.

You'll find instructions on how to build on the [wiki (Building from source)](https://github.com/la5nta/pat/wiki/Building-from-source). You will need to run `go generate` to generate the new web assets bundle after an initial `./make.bash`, and then re-run `./make.bash` and you should see the changes :)

Thanks!

res/tmpl/index.html
```diff
@@ -35,19 +35,15 @@
             <li><a href="#" id="outbox_tab">Outbox</a></li>
             <li><a href="#" id="sent_tab">Sent</a></li>
             <li><a href="#" id="archive_tab">Archive</a></li>
-            <li class="dropdown">
+<!--          Flattened menu as the additional dropdown was annoying
```
DC7IA, line 38: The comment is not needed when merged, I think this should be removed.

DC7IA:
> Thanks for helping out Joshua +1

I'm glad I found a way to help.

Also, the reason it did not compile was simply the fact the folder was not named la5nta/pat, but bitmux/pat... Had to rename it, which took a while to figure out.

This is how it looks now:

Screenshot_2020-05-28_00-14-54


I think this is an improvement to before. :)

And I also think we should put more icons in there.
I guess there are similar icons that can be found for Inbox, Outbox and Archive. 
That way it will look even more clean and more user friendly, as it breaks up the line of text and makes it visually more appealing. And easier to locate the right button.

DC7IA:
Now that we're talking about the dropdown menus...

Is there any reason for this menu?

Screenshot_2020-05-28_00-24-40

Does it need to be there? We already have the other menu in the same dialog:

Screenshot_2020-05-28_00-25-24

I do not even know what the difference is. 

(One of the confusing things I talked about in #156.)

martinhpedersen:
Regarding the "Aliases" menu:

This is a feature where users can add their favorite nodes in the config file for easier access. The "telnet" alias is simply the configuration needed to connect to Winlink's CMS over the Internet, but you can extend that list with your favorite ARDOP/WINMOR/AX.25/... node (with frequency). Mine looks like this:

image

martinhpedersen:
Thank you for providing screenshots @DC7IA 😄  I see what you mean regarding more icons, if we do decide to flatten the menu.

What really concerns me is how this will behave on smaller screens, and phones in portait orientation...

Maybe we should post some screenshots showing this design on various devices @ pat-users, to get some feedback from the community?

DC7IA:
Well, I'd say the web frontend is very responsive:

Screenshot_2020-05-28 Pat - Mailbox

This is how it looks with the menu closed:

Screenshot_2020-05-28 Pat - Mailbox_menu_closed

And in vertical orientation:

Screenshot_2020-05-28 Pat - Mailbox_vertical

martinhpedersen:
Aah, thanks! I've forgotten how responsive that navbar is 😆 

I think this looks promising. If we could add some icons for the mailboxes, and/or do something to visualize the difference between the "actions" and the mailboxes I really think you're on to something here.

Maybe we could pull the "action items" to the far right?
image
(imagine the search bar is where we put the actions..

What do you think?

DC7IA:
> Maybe we could pull the "action items" to the far right?

I think it would look inconsistent:

On bigger screens there'd be a big gap and on smaller screens (like tablets) no gap at all.

I think it should look as similar as possible on all devices to look consistent.

Some simulations using firefox developer console:

Nexus 7

Screenshot_2020-05-28 Pat - Mailbox_Nexus7

Nexus 10

Screenshot_2020-05-28 Pat - Mailbox_Nexus10

iPad

Screenshot_2020-05-28 Pat - Mailbox_iPad

martinhpedersen:
Thanks for the screenshots @DC7IA, I think these provide a decent overview of the proposed change 👍 

I'm not sure I agree with you regarding the right alignment of actions though.. I see what you mean, but I'm not sure it will be a big problem. On large ("normal") screens we would have a clear separation between mailboxes and actions, on phones it does not matter (collapsed menu) and on the mid range we would have non (or less) separation.

Actually, it's used as one of the example on [w3schools's Bootstrap Navbar page](https://www.w3schools.com/bootstrap/bootstrap_navbar.asp), to separate "actions" from "pages":

image

I'm open for other suggestions, but I believe we would benefit from some sort of separation between the two categories.

DC7IA:
> Actually, it's used as one of the example on [w3schools's Bootstrap Navbar page](https://www.w3schools.com/bootstrap/bootstrap_navbar.asp), to separate "actions" from "pages":
> 
> image

Yeah, that looks better. I suggest placing "About Pat" at the bottom to have less buttons at the top. Because that will probably be the least used one.

martinhpedersen:
👍 

How about some sort of middleway here? Move the more commonly used actions out to the navbar (possibly only *Connect...*), but keep the "actions" dropdown for less used actions (rename to "more"?). That would also scale better if we in the future add even more actions.

That would certainly give more space between "mailboxes" and "actions" if we do decided to to use that approach.

DC7IA:
I'd prefer changing the compose dialogue to one that lets you switch to position mode, that way we'd only have "Compose", not "Compose" and "Position", because typing in a location is actually composing a message as well.

This is how it would look:
   
Screenshot_2020-05-28 Pat - Mailbox_edited

I think we should then split up the compose box into two tabs, one for text messages and one for positions.

What do you think?

DC7IA:
I made a quick mock-up of what I imagine:

Screenshot_2020-05-28_21-36-42

Of course with a different background colour, this isn't really readable, but the best I could do within a few minutes.

martinhpedersen:
I like the idea! Also, a tab based modal for Connect would also be useful, when we finally implement the rmslist feature for the web gui.

It does not resolve the issue regarding future proofing though, as we will be limited to a rather small amount of actions with this design.

I'm also afraid to feature creep on this. Merging these modals would require a rather large re-write of some major parts of the GUI, and I believe some of the other features and issues needs to be adressed first.

I still think keeping a dropdown menu ("more") for the less common actions would be beneficial until we have time to merge some of these modals into a deeper view hierarchy.

How about adding some icon-only "action shortcuts" on the right side (compose, position, connect). We could then keep the drop-down menu (maybe as a ⋮ icon) with the full set of actions. That way we would give the user easy access to these functions, but still have a place to stash less frequently used actions. (I guess we need to draw this in order to fully visualize it :p)

Thank you for discussing this with me by the way :)

DC7IA:
> How about adding some icon-only "action shortcuts" on the right side (compose, position, connect). We could then keep the drop-down menu (maybe as a ⋮ icon) with the full set of actions. That way we would give the user easy access to these functions, but still have a place to stash less frequently used actions. (I guess we need to draw this in order to fully visualize it :p)

Maybe then use tooltips for the icon-only actions as requested in https://github.com/la5nta/pat/issues/156#issuecomment-540723103_

Otherwise it could become pretty confusing if you cannot lookup the meaning of a button.

What I don't understand:

>It does not resolve the issue regarding future proofing though, as we will be limited to a rather small amount of actions with this design.

Could you explain how it is limiting? I mean, there could be many different tabs, Right now we'd only need two and I doubt we will ever have a thousand different actions.

Right now we have messages and positions and maybe weather reports in the future..? But that is all I can think of. (Airmail, as an example, can request news from different categories.)

Another idea I had was a big form with different parts you choose from with radio buttons:

```
(x) message

from:
to:
message:

(...)

( ) position
position:
```

something like that, but I think tabs would be better. Because this could be a really big form in the end, not as easy to use as tabs, I think.

res/tmpl/index.html
bitmux: Fair enough

Issue #122, opened 2018-03-15
Drop support for Go 1.6


martinhpedersen:
Go 1.6 is [ancient](https://golang.org/doc/devel/release.html#policy) and stopped receiving security patches several years ago.

The reason to keep supporting it was to ensure that Pat could be compiled with the Go version available in debian stable backports.

Debian Stretch backports have Go 1.8 now.. so I guess that is a reasonable new minimum Go version.

kitchen:
raspbian seems to ship with go 1.7 at least as of a few days ago when I got this all up and running.

martinhpedersen:
Debian Buster (latest stable) includes Go 1.11, so I guess that would be a reasonable new minimum version.

guitarpicva:
My Linux go version is 1.10.4 in a current Mint version, via the pkg mgr.  Mint is a very popular Ham platform.

Is the go upgrade process relatively painless?  I haven't tried it yet.

martinhpedersen:
Thanks for the info :) I ended up keeping support for 1.10, so it should work just fine with the Go version shipped with Mint on rpi :+1: 

It's very easy install the official releases from golang.org actually. It's as simple as extracing a tarball and updating your PATH. It's the recommended way of installing Go, to get the latest features and fixes.

Instructions here if you want to give it a Go ;-) https://golang.org/doc/install#tarball

guitarpicva:
Many/most are package hounds only however.  Keeping 1.10 for now if probably wise.

PR #199, opened 2020-05-27
Manage Go dependencies with Go Modules

martinhpedersen:
Switching to [Go Modules](https://blog.golang.org/using-go-modules) for managing Go dependencies for this project.

Go Modules has been the recommended dependency manager for some years now, and is more integrated with the toolchain than using git submodules.

We will need to update the [Building from source wiki page](https://github.com/la5nta/pat/wiki/Building-from-source) to reflect these changes. Should also add a section on how to build with local modules.

This changes the minimum Go version to 1.11 (ref #122)


harenber:
@martinhpedersen I would need github.com/albenik/go-serial/v2 for the SCS driver. Would you handle that or to you want me to prepare a pull request against the "develop" branch?

PR #91, opened 2017-08-16
Man page

Andrew-H3:
Hey,
It's Andrew. I started working on the man page for pat. I'm not sure how it works with packaging, but for now I just put it in the debian directory. It's basically empty for now, but I'll work on it as I can and try to get most of it done soon.

PR #203, opened 2020-06-05
Fixes typo and title of pat manpage

taowa:


martinhpedersen:
Thanks for this @taowa.

I would really like to keep the version number out of the man page though, to avoid having even more sources to update when bumping the version prior to new releases. We could also look into auto-generating the man page as part of the debian packaging.

Most of the content of this man page was copied from the output of `pat help`. Do you mind checking if the typo is present there as well?

Thank you!

martinhpedersen:
Ping @taowa

martinhpedersen:
No response from author. Closing for now.

taowa:
Whoops, sorry, I don't check github all that often. The typo was fixed in commit 0acc993ef34fc832d61419630a0bfec1258bfbc9. This probably means the man page was last generated some time before then. Do you remember how it was generated? I see no issues in principle with doing so as part of the packaging as long as it's not generated by doing something too weird.

martinhpedersen:
No problem @taowa 🙂 

I'm afraid the current man page(s) was written by hand. See PR #91. IMHO the current man pages should be replaced by something auto-generated or removed, unless we are able to find someone willing to maintain them properly.

What do you thing? Do you have any insights as to how these things are managed in other (similar) projects? Any tools we could use?

martinhpedersen:
I realize this is way of topic for the changes you requested. I'd be happy to merge these changes as you provided them. But my main concern is having the version number there. Can we do without it?

Thanks!

taowa:
Martin Hebnes Pedersen, 2020-11-11 15:57 -0500:
> I'm afraid the current man page(s) was written by hand. See PR #91.
> IMHO the current man pages should be replaced by something
> auto-generated or removed, unless we are able to find someone willing
> to maintain them properly.
>
> What do you thing? Do you have any insights as to how these things are
> managed in other (similar) projects? Any tools we could use?

help2man is... alright :). Briefly tested it, seems to output an extra
copy of the "commands" section for some reason with pat. I'll look into
why.

Martin Hebnes Pedersen, 2020-11-11 16:23 -0500:
> I realize this is way of topic for the changes you requested. I'd be
> happy to merge these changes as you provided them. But my main concern
> is having the version number there. Can we do without it?

I'm partial to the version number being equal to the version for which
the man page represents and not the most recent version. In this case,
though, it's likely not 0.9.0, so it may be more appropriate to change
the topic of the PR (or close it and make it into an issue). Benefits
include that if a new man page isn't generated for some reason, a user
could possibly determine that the mismatch between the version of the
manpage and version of pat might be why things don't work.

Then again, considering that I've never looked at the versions on my
manpages, this might be wishful thinking ;).

I'll take a look at what's up with help2man and try to reply with
details and what might be the easiest way to generate the manpage when I
have a moment. Might be a little while though.

Taowa

-- 
Taowa (they)
taowa.ca
LOC FN35EL

martinhpedersen:
The typo was fixed in #313
 The second change (version in man file) is something I would rather avoid, since it's yet another place to remember to update.

Closing this now. But thanks for the PR and the good discussions 🙂 I hope we can improve the man pages down the road.

PR #208, opened 2020-07-26
Add the tmd710-tncsetup script as a suggest

fmarier:
The script that is referenced in `/etc/default/ax25` is now packaged for Debian: https://packages.debian.org/bullseye/tmd710-tncsetup


martinhpedersen:
This is great news @fmarier! Thank you 👍 

I've rebased and merged to `develop`. Aiming to include this in the next release.

PR #226, opened 2020-10-22
Add rig_mode parameter to URL

DougCollinge:
Add a parameter 'rig_mode' that sets the modulation mode of the current VFO
of the rig. The previous mode is preserved and restored later. An input on
the 'Connect' panel shows the rig_mode and allows it to be changed prior to
a connect. 

Currently, this probably only works with rigctld, not the hamlib API.


DougCollinge:
I'm pretty new at practically everything involved in this PR so please be gentle with me! Tested only on Linux at this point.

DougCollinge:
I found a bug already! The QSY that happens when you change something in the connect form does not change the rig mode, just the frequency. I'll work on that and update the PR (if I can figure out how!)

DougCollinge:
I have a better patch in progress.

PR #230, opened 2020-10-26
fixed #229 added support for azimuth NaN, which happens if distance is 0

### LINKED ISSUE #229: RMS list fails to populate when distance is 0, azimuth is NaN
The JSON serializer in rmslistHandler fails silently when one of the RMSes has a distance of 0 and azimuth is NaN. In that case the RMS list display on the connect dialog does not get updated.

rainerg2000:
For locator CN85pl the RMS list list on the connect dialog was not working correctly for Packet @ 2m. 
For K7GDS-10 the distance is 0 and azimuth is NaN. The JSON serializer couldn't deal with NaN.

### DIFF
--- a/http.go
+++ b/http.go
@@ -386,7 +386,11 @@ func rmslistHandler(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 	sort.Sort(byDist(list))
-	json.NewEncoder(w).Encode(list)
+	err = json.NewEncoder(w).Encode(list)
+	if err != nil {
+		log.Println(err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+	}
 }
 
 func qsyHandler(w http.ResponseWriter, req *http.Request) {
--- a/rmslist.go
+++ b/rmslist.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"math"
 	"net/url"
 	"path"
 	"strings"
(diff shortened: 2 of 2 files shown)


martinhpedersen (changes requested):
Nice catch 👍 😄 

But since this is related to the JSON serialization, I would rather we use a custom float64 type (JSONFloat64) for those fields to make them serialize as `null` if NaN (or Inf). I believe that would be a better approach with regards to [SoC](https://en.wikipedia.org/wiki/Separation_of_concerns).

Somethink like this (not tested):

```go
// JSONFloat64 is a float64 which serializes NaN and Inf(+-) as JSON value null
type JSONFloat64 float64

func (f JSONFloat64) MarshalJSON() ([]byte, error) {
  if math.IsNaN(float64(f) || math.IsInf(float64(f), 0) {
    return json.Marshal(nil)
  }
  return json.Marshal(float64(f))
} 

type RMS struct {
  ...
  Distance JSONFloat64 `json:"distance"`
  Azimuth  JSONFloat64 `json:"azimuth"`
  ...
}
```

Thoughts? :)

rainerg2000:
sounds good. I'll try that.

rainerg2000:
done

martinhpedersen (approved):
Perfect 👍  Thanks Rainer!

martinhpedersen (commented):
Do you mind squashing these changes into a single commit and rebasing on top of `develop`?

rainerg2000:
squashed and rebased to la5nta/develop. I just realized that this PR merges to la5nta/master. 
@martinhpedersen, that is probably not what you'd want. Should I create another PR to merge to develop ?

martinhpedersen:
Thanks Rainer🙂  No need to close the PR to change base branch though! I fixed it 👍

PR #232, opened 2020-11-14
Improve results of message 'reply' and 'forward' buttons

DougCollinge:
Make 'reply' open the form with two blank lines followed by the quoted reply
text, and position the text input cursor at the top, same as other email clients.
Make 'forward' fill the form with the quoted text, no blank lines, and clear 
and put the text input cursor into the 'To' field, since you are presumably 
forwarding it to someone else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ Copyright (c) 2020 Martin Hebnes Pedersen LA5NTA
 * LA4TTA - Erlend Grimseid
 * LA5NTA - Martin Hebnes Pedersen
 * W6IPA  - JC Martin
+* VE7GNU - Doug Collinge
 
 ## Thanks to
 
--- a/res/js/index.js
+++ b/res/js/index.js
@@ -897,18 +897,23 @@ function displayMessage(elem) {
 			} else {
 				$('#msg_subject').val(data.Subject);
 			}
-			$('#msg_body').val(quoteMsg(data));
+			$('#msg_body').val('\n\n'+quoteMsg(data));
 
 			$('#composer').modal('show');
+			$('#msg_body').focus();
+			$('#msg_body')[0].setSelectionRange(0,0);
+			
 		});
 		$('#forward_btn').off('click');
 		$('#forward_btn').click(function(evt){
 			$('#message_view').modal('hide');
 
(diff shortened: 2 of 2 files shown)


martinhpedersen (approved):
This is excellent Doug 👍 

Do you mind adding yourself to the Contributors list in `README.md`? You can just do it as a secondary commit on this branch. Thanks!

DougCollinge:
Ok, It seems I forgot to push a commit but it's done now. I'm still fumbling with git so I didn't have the courage to attempt to squash those two commits together.

And I added myself to the README - thanks! It's such a small contribution...

martinhpedersen (changes requested):
Thanks 👍 

Despite being small changes, these little details are great contributions. And as an added bonus, small PRs are very easy to review and merge 🙂 

The `.gitignore` changes should be omitted though, especially `bindata_assetfs.go` which is an important file that should not be ignore by git. The pat executable should probably be in gitignore, but I'd prefer to keep PRs as specific as possible to make them easier to review.

If you prefer, I can fix this for you by cherry-picking all but the first commit. But it's a nice opportunity for you to improve your [git rebase](https://medium.com/@daendersby/til-git-rebasing-removing-rage-commits-4b5c42e7af3a) skills 🙂 

Thanks again!

DougCollinge:
Thanks for the opportunity to improve! Here's an update with the changes you requested. (I think)

martinhpedersen (changes requested):
Hmm.. it seems like you updated your branch, and then merged in the old version. So this branch now contains 7 commits, but should really only contain two.

After rebasing your branch, you need to do a "force push" to github. I suspect you attempted a normal push, got an error, pulled and then pushed again.

1. Do an interactive rebase and drop all but the two relevant commits (80548621d0b115ae7fb6e274c08f5ffb2b072346 and de4d22edcf0ab68e4627378057c188b3916121da).
2. `git push -f`

😄

martinhpedersen:
... alternativly you could start with a clean branch of `develop` and cherry-pick the relevant commits. But the interactive rebase is very powerful, so I would focus on learning that 🙂

martinhpedersen (approved):
Perfect! Thank you 😊

PR #235, opened 2020-11-16
Fix inappropriate alpha sort on distance and frequency in RMS list

DougCollinge:
Use a lurid hack to fix this problem by right justifying the distance and frequency strings by prepending enough Unicode NO-BREAK SPACE characters \u00A0. The header-click sort function would normally trim leading spaces but is fooled by these non-spaces and doesn't trim them. Thus the resulting alpha sort also sorts numerically!

### DIFF
--- a/freq.go
+++ b/freq.go
@@ -46,7 +46,6 @@ type Frequency int // Hz
 func (f Frequency) String() string {
 	m := f / 1e6
 	k := (float64(f) - float64(m)*1e6) / 1e3
-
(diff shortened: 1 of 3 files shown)


DougCollinge:
I have no idea what Travis is griping about...

xylo04:
Hey Doug, it's an impressively creative workaround, but you're right about it being a little gnarly. There's got to be a better way, right?

PR #237, opened 2020-11-24
Allow Unsorted Connect Aliases

DougCollinge:
Connect aliases are presented sorted into alphanumeric order, which is unnecessary and may be undesired. If the aliases are presented in the order they are entered in config.json then they can be in any order the user desires. This patch allows an unsorted ordering of connect aliases.

The re-ordering comes from storage in a Go map, which has no defined ordering. This patch uses  an array for the same purpose, which prevents re-ordering because arrays are inherently ordered.

In order to maintain backward compatibility in the format of config.json an optional new structure is allowed named unsorted_aliases of type [][]string. The first element of the subarray is the alias key and the second element is the URL. If entries are present in unsorted_aliases then they are presented in the aliases dropdown in unsorted order followed by any entries in connect_aliases sorted in alpha order.

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -77,6 +77,13 @@ type Config struct {
 	// Any occurrence of the substring "{mycall}" will be replaced with user's callsign.
 	ConnectAliases map[string]string `json:"connect_aliases"`
 
+	UnsortedAliases [][]string `json:"unsorted_aliases"` 
+
+	AllAliases struct {
+		ConnectAliases *map[string]string `json:"connect_aliases"`
+		UnsortedAliases *[][]string `json:"unsorted_aliases"`
+	} `json:"all_aliases"`
+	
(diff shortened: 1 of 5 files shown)


martinhpedersen:
I think you are on to something here👍 Although it would be a more verbose config syntax,an ordered aliases list seems reasonable.

I'm not sure I agree with the proposed data types though. How about defining a new type Alias instead of the inner-most level and make it a `[]Alias`?

```go
type Alias struct {
  Name string `json:"name"`
  URL  sring `json"uri"`
}
```

Also, with some customer json.Unmarshaller implementiation we can maintain backwards compatibility with the old format without having to introduce a new config parameter 🙂

martinhpedersen:
@DougCollinge - Any chance you're interested in working more on this PR? 🙂

PR #242, opened 2021-02-05
Update go deps

xylo04:
No strong reason, just housekeeping.


xylo04:
Afflicted by https://github.com/golang/go/issues/42665, need to upgrade golang version.

PR #244, opened 2021-03-19
Attempt to parse non-multipart form too

xylo04:
The web frontend sends composed message with content type `multipart/form-data`, but some folks are now experimenting with scripting against the API. They ought to be able to send as `application/x-www-form-urlencoded` when their libraries use it.

### DIFF
--- a/http.go
+++ b/http.go
@@ -9,10 +9,12 @@ import (
 	"bytes"
 	"embed"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"html/template"
 	"io"
 	"log"
+	"mime/multipart"
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


martinhpedersen:
Thanks Chris!

Looking at the documentation of ParseMultiForm, it looks like ParseForm is already called indirectly:
> ParseMultipartForm calls ParseForm automatically. ParseForm is idempotent.

Have you investigated this?

xylo04:
Rebased the PR, but I'm still investigating. I do recall seeing in the library code that ParseMultiForm calls ParseForm, but for some reason I don't think that was working as expected...

xylo04:
Ok, so I was able to create curl requests that reproduce the issue, and debugging helped me understand the issue.

This request works prior to this PR:
```
curl -v --location --request POST 'http://localhost:8080/api/mailbox/out' \
--form 'date="2021-06-15T00:30:57Z"' \
--form 'subject="Hello"' \
--form 'to="K0SWE-2"' \
--form 'body="This is a test of multipart form-data."'
```

But this one doesn't work before the PR:
```
curl -v --location --request POST 'http://localhost:8080/api/mailbox/out' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--data-urlencode 'date=2021-06-15T00:32:51Z' \
--data-urlencode 'subject=Hello' \
--data-urlencode 'to=K0SWE-1' \
--data-urlencode 'body=This is a test of x-www-form-urlencoded.'
```

The reason is that `http.Request.ParseForm()` takes the request body and populates the `http.Request.Form` field which is necessary for both content types, but `http.Request.ParseMultipartForm()` still forces the content type to be multipart.

However, going back and looking at this again makes me realize the PR doesn't work as-is because the the rest of the creation request only uses the `MultipartForm` field. I'll do some additional work on this.

xylo04:
Ready for another review.

http.go
```diff
@@ -302,6 +265,54 @@ func postOutboundMessageHandler(w http.ResponseWriter, r *http.Request) {
 	fmt.Fprintf(w, "Message posted (%.2f kB)", float64(buf.Len()/1024))
 }
 
+func attachFile(w http.ResponseWriter, f *multipart.FileHeader, msg *fbb.Message) error {
```
martinhpedersen, line 268: I really like that you moved this out to separate function. It really cleans up the handler 🙂 

I think it would be even better if we could remove the ResponseWriter argument, to make it both less verbose and get rid of the HTTP response details in this function.

Of course, we still need to support returning different status codes based on the nature of the errors. One approach I've seen (and used) multiple times, is by defining a custom error type that wraps the original error and adds a StatusCode field.

https://play.golang.org/p/FxGw97vp2Gr

How do you feel about that?

martinhpedersen (commented):
Thanks for taking the time to investigate it further. I agree with you in that this would be a great improvement for those wanting to call the endpoint from a bash script 👍  

The changes overall looks very nice. I only have one concern regarding the new `attachFile` function. See comment.

http.go
xylo04: Done!

martinhpedersen (approved):
Perfect!  Thank you 😄

martinhpedersen (changes requested):
Oh, I missed one minor detail 😊 

`HttpError` -> `HTTPError`

It is more idiomatic (for Go code) to use consistent casing of acronyms and initialisms. Either all caps or no caps. You'll see it all over the place in the Go standard library as well as most other projects: [json.HTMLEscape](https://golang.org/pkg/encoding/json/#HTMLEscape), [http.Handler's ServeHTTP-method](https://golang.org/pkg/net/http/#Handler).

golint actually complains about this (amongst lot of other things):
> http.go:67:6: type HttpError should be HTTPError

More on that subject here, if you're interested: https://github.com/golang/go/wiki/CodeReviewComments#initialisms.

Sorry I didn't catch that on the previous review.

xylo04:
Done! Bad Java habits I need to break. 😉

martinhpedersen (approved):
That's perfect. Thank you 🥇😃

PR #250, opened 2021-05-30
Update go deps

xylo04:
No strong reason, just housekeeping.

### DIFF
--- a/http.go
+++ b/http.go
@@ -83,6 +83,10 @@ func ListenAndServe(addr string) error {
 	r := mux.NewRouter()
 	r.HandleFunc("/api/connect_aliases", connectAliasesHandler).Methods("GET")
 	r.HandleFunc("/api/connect", ConnectHandler)
+	r.HandleFunc("/api/formcatalog", formsMgr.GetFormsCatalogHandler).Methods("GET")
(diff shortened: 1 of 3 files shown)


xylo04:
By the way, let me know if you'd rather do these one PR at a time.

martinhpedersen:
Thanks!

My main concern about merging this is that we might end up breaking some of the less used features, or something that I'm not able to test with my current working conditions (e.g. ax25 and/or pactor connections).

Up until now, I have manually reviewed (more or less) all changes in the dependencies before upgrading them.

If only we had a group of users dedicated to testing the `develop` branch... or even better, some sort of alpha channel to ship automatic builds based on `develop` to a group of testers.

So I guess my question is; Have you reviewed (or tested) these changes enough that you're confident we don't break something by releasing this? 🙂 Keep in mind that the release process is time consuming, so I really don't want to release based on a bad branch.

... We really need to improve and simplify the release process. (let's continue that discussion in the ongoing email thread).

xylo04:
Yeah, it's true that I hadn't tested thoroughly, I sometime do things things sort of blindly relying on being able to roll back. I'll withdraw for now until we can come up with a safer way.

PR #255, opened 2021-06-20
Don't use converted images if they're bigger than original

### LINKED ISSUE #243: Attachments shall not be recompressed when the result file size is larger than the original.
### Tested version
Pat v0.10.0 (601f6d3) linux/amd64 - go1.15.1

### Reproduction path
- Run pat with the web server.
- On the web page, select "Action" > "Compose".
- Write an e-mail address and a subject
- Attach the following picture: 

- Send the mail
- Open the mail in the outbox
- Download the attachment
- Check the file size

### Actual Behaviour
When attaching pictures to a mail, pat attempts to re-compress the file. 
Pat attaches the re-compressed file, even when the original file is smaller. 
Test vector size: 1.8KB
Attached file size: 3.5KB

### Expected Behaviour
When attaching pictures to a mail, pat attempts to re-compress the file. 
Pat attaches the original file when it is smaller. 
Test vector size: 1.8KB
Attached file size: 1.8KB

xylo04:
Fixes #243. Confirmed with manual test.

<a href="https://ibb.co/NKgnrF9">Screenshot-from-2021-06-20-17-18-55</a>

### DIFF
--- a/convert_image.go
+++ b/convert_image.go
@@ -8,7 +8,6 @@ import (
 	"bytes"
 	"image"
 	"image/jpeg"
-	"io"
 	"mime"
 	"path"
 	"strings"
@@ -31,8 +30,8 @@ func isImageMediaType(filename, contentType string) bool {
 	return strings.HasPrefix(mediaType, "image/")
 }
 
-func convertImage(r io.Reader) ([]byte, error) {
-	img, _, err := image.Decode(r)
+func convertImage(orig []byte) ([]byte, error) {
+	img, _, err := image.Decode(bytes.NewReader(orig))
 	if err != nil {
 		return nil, err
 	}
@@ -44,6 +43,11 @@ func convertImage(r io.Reader) ([]byte, error) {
 
 	// Re-encode as low quality jpeg
 	var buf bytes.Buffer
-	err = jpeg.Encode(&buf, img, &jpeg.Options{Quality: 40})
-	return buf.Bytes(), err
+	if err := jpeg.Encode(&buf, img, &jpeg.Options{Quality: 40}); err != nil {
+		return orig, err
+	}
+	if buf.Len() >= len(orig) {
+		return orig, nil
+	}
+	return buf.Bytes(), nil
 }
--- a/http.go
+++ b/http.go
@@ -313,7 +313,7 @@ func attachFile(f *multipart.FileHeader, msg *fbb.Message) error {
 	if isImageMediaType(f.Filename, f.Header.Get("Content-Type")) {
 		log.Printf("Auto converting '%s' [%s]...", f.Filename, f.Header.Get("Content-Type"))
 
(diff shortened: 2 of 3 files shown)


http.go
```diff
@@ -315,11 +315,15 @@ func attachFile(f *multipart.FileHeader, msg *fbb.Message) error {
 		if converted, err := convertImage(bytes.NewReader(p)); err != nil {
 			log.Printf("Error converting image: %s", err)
 		} else {
-			log.Printf("Done converting '%s'.", f.Filename)
+			if int64(len(converted)) < f.Size {
```
martinhpedersen, line 318: That's nested a little deeper than I prefer. How about refactoring so that `func convertImage()` becomes a no-op if the converted image is larger than the original?

martinhpedersen (changes requested):
Thanks for having a go at this. It's been bugging me for some years now 😄 

See inline comment.

http.go
xylo04: Done

xylo04:
Woops, that doesn't build...

xylo04:
Ready for another review.

convert_image.go
```diff
@@ -45,5 +44,9 @@ func convertImage(r io.Reader) ([]byte, error) {
 	// Re-encode as low quality jpeg
 	var buf bytes.Buffer
 	err = jpeg.Encode(&buf, img, &jpeg.Options{Quality: 40})
-	return buf.Bytes(), err
+	converted := buf.Bytes()
+	if len(converted) < len(orig) {
+		return converted, err
+	}
+	return orig, err
```
martinhpedersen, line 51: I try my best to follow the [Return Early pattern](https://medium.com/swlh/return-early-pattern-3d18a41bba8) in this project, as it is preferred in most Go codebases.

One of the key points of this pattern is the [Happy Path](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88), which I don't think is followed here.

Following that pattern plus the convenient `err := doStuff(); err != nil` syntax and using `Buffer.Len()`, I get this code:

```go
if err := jpeg.Encode(&buf, img, &jpeg.Options{Quality: 40}); err != nil {
	return orig, err
}
converted := buf.Bytes()
if buf.Len() >= len(orig) {
	return orig, nil
}
return buf.Bytes(), nil
```

I think that's more readable as we here explicitly handle the error right away by returning early, and can clearly see the happy path and exception handling by following the indent level.

I'd prefer if you change that please 🙂 

PS: You can go ahead and squash the commits afterwards, and I'll merge it asap. Thanks!

martinhpedersen (changes requested):
See inline comment 🙂 

And, again thank you for this!

convert_image.go
xylo04: Done and squashed

martinhpedersen (approved):
Perfect! Thank you 😀

PR #257, opened 2021-06-23
Configure a default forms path

### LINKED ISSUE #256: Add a good default value for forms_path
We should add a good default value for forms_path, so that no configuration is needed to get started with html forms.

Since all other state is stored under the .wl2k directory, I think that is a good place to put it.

Similar to fConfigPath, fLogPath, fMailboxPath... it should probably be defined as a CLI option (fOptions) with a default value relative to appDir rather than a config parameter. (See main.go)

@xylo04 @rainerg2000 - What do you think?

Relates to #254, #233 and #234.

xylo04:
This primes the user to run a forms update with zero configuration, but doesn't actually do it for them.  This could cause errors if they try to use forms before running the update. With some more work, we could make this more like the RMS list where the update happens proactively. For now, I think this is good enough.

Fixes #256

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -6,6 +6,8 @@ package cfg
 
 import (
 	"encoding/json"
+	"github.com/la5nta/wl2k-go/mailbox"
+	"path"
 	"strings"
 
 	"github.com/la5nta/wl2k-go/transport/ardop"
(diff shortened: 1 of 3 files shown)


martinhpedersen (approved):
Thanks :)

PR #260, opened 2021-07-07
Bump version (0.11.0)

xylo04:


martinhpedersen:
Thanks, but I already have this commit prepared in a private repo 😊 See comment in the google doc.

By the time you wrote this, I had already prepared the changelog entries and have started to package the binaries, and if I where to merge this instead of my own local branch I would have to rebuild all those binaries. Sorry I did not notify you in advance.

Your work looks very good though, so thank you for the effort😊 Next time, I'll be happy to involve you more in the release process👍 Thanks!

xylo04:
No problem, sounds good

PR #264, opened 2021-07-16
Add http to log message

xylo04:
This very slight change makes the serving address a clickable link in many terminal emulators, which is a nice qualtiy-of-life improvement.


martinhpedersen (approved):
Great idea 👍

PR #265, opened 2021-07-16
Address most linter errors in http.go

xylo04:


martinhpedersen (approved):
Thanks. It goes to show that we need to implement better error handling in the http handlers. I have a pattern in mind, so I'll fix that when I have time 👍

PR #268, opened 2021-07-22
Add basic debug logging capabilities

martinhpedersen:
@xylo04 - Ref email discussion. How about this? 🙂 

By implementing it as a separate package like this, we'll be able to call it from within the forms package and any of the other internal packages.

### DIFF
--- a/internal/debug/debug.go
+++ b/internal/debug/debug.go
@@ -1,0 +1,25 @@
+package debug
+
+import (
+	"log"
+	"os"
+	"strconv"
+)
+
+const (
+	EnvVar = "PAT_DEBUG"
+	Prefix = "[DEBUG] "
+)
+
+var enabled bool
+
+func init() {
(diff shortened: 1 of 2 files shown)


internal/debug/debug.go
```diff
@@ -0,0 +1,24 @@
...
+	"strconv"
+)
+
+const (
+	EnvVar = "PAT_DEBUG"
+	Prefix = "[DEBUG] "
+)
+
+func enabled() bool {
+	b, _ := strconv.ParseBool(os.Getenv(EnvVar))
```
xylo04, line 15: This probably isn't that expensive, but it's potentially called many times. Is it worth initializing a variable for this?

xylo04:
Looks good otherwise!

internal/debug/debug.go
martinhpedersen: I agree 👍  No use in supporting env variable changes while the application is running.

I've fixed it by parsing the env variable in func init() 🙂

PR #271, opened 2021-08-02
Start using Prettier formatter for web resources

xylo04:
The https://prettier.io/ formatter is popular for web projects. This change looks large, but really it's whitespace and quote type changes in the existing files, plus configuration to install the formatter tool itself.

Typical usage after clone would be to `cd` into the `res` directory, `npm install` to download the tool, and `npm run format` to execute the formatter.

If you want to suggest changes to the style in `res/.prettierrc` I'm open to that, but bottom line, I'd like to see the web resource become consistently formatted. What do you think?


martinhpedersen:
I really like this 👍 

My only concern is that everything inside the res/ folder is [embedded](https://github.com/la5nta/pat/blob/master/http.go#L38) in the binary at build time, but `node_modules` and these new config files should not be included.

How about moving `/res` to `/web/res` and changing the embed code to only include that sub-directory? We could then keep node_modules and all the config at `/web/`, and avoid embedding those files.

-----

Side note: It would be nice if we added a CI step to ensure that all code has been formatted correctly by running both go fmt (even better: [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports)) and prettier. But let's discuss that elsewhere 😊

xylo04:
> everything inside the res/ folder is [embedded](https://github.com/la5nta/pat/blob/master/http.go#L38) in the binary at build time, but `node_modules` and these new config files should not be included.

Ah, good point, will do.

xylo04:
Moved the web resources as you suggested.

http.go
```diff
@@ -72,7 +74,12 @@ type HTTPError struct {
 var websocketHub *WSHub
 
 func ListenAndServe(addr string) error {
+	var err error
 	log.Printf("Starting HTTP service (http://%s)...", addr)
+	staticContent, err = fs.Sub(embeddedFS, "web")
+	if err != nil {
+		log.Fatal(err)
+	}
```
martinhpedersen, line 82: Since this loads embedded content, maybe we could do it `func init()` (above func ListenAndServe)? .. possibly also panic'ing on any error.

```go
func init() {
  var err error
  staticContent, err = fs.Sub(embeddedFS, "web")
  if err != nil {
    panic(err)
  }
}

func ListenAndServe(addr string) error {
...
```

(Otherwise I would prefer that we return the error from fs.Sub instead of calling log.Fatal, since ListenAndServe already returns an error.)

http.go
```diff
@@ -343,7 +350,10 @@ func wsHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func uiHandler(w http.ResponseWriter, _ *http.Request) {
-	data, err := staticContent.ReadFile(path.Join("res", "tmpl", "index.html"))
+	f, _ := staticContent.Open(path.Join("res", "tmpl", "index.html"))
+	fInfo, _ := f.Stat()
+	var data = make([]byte, fInfo.Size())
+	_, err := f.Read(data)
```
martinhpedersen, line 356: Is it possible to simplify this by using [`fs.ReadFile()`](https://pkg.go.dev/io/fs#ReadFile)? It should behave much like the old ReadFile method, only you will need to pass in staticContent as the first parameter 🙂

martinhpedersen (changes requested):

http.go
xylo04: Done

http.go
xylo04: Ah, that's much nicer. I tried doing staticContent.ReadFile() but it wasn't on the interface, thanks for pointing this out.

martinhpedersen (approved):
Thank you 😊

PR #277, opened 2021-08-21
Upgrade jQuery from v1.11.1 to v3.6.0

xylo04:
I did some manual testing and didn't see issues. Let me know if there are specific things you would like to test.


xylo04:
Martin, any concerns?

martinhpedersen:
The major version change v1 -> v3 suggests that this upgrade involves breaking changes. I think we need to check carefully that we're not affected by any of these.

Have you looked at the changelog and/or any migration guide?

I think we either should do this, or manually test most of the app use-cases while carefully examining the js console output.

Suggestions for manual testing (from the top of my head):
* Are any of the API calls (including error handling) affected?
* Any changes related to how we use the websocket?
* Is the attachment preview (in particular images) behaving correctly?
* Any changes affecting the status-panel?

xylo04:
There's documentation at https://jquery.com/upgrade-guide/, I'll start looking through that.

xylo04:
I tested .show(), .hide(), and .toggle(), and all of the modal dialogs and popovers seem to behave appropriately. Also, I tested image upload by sending you a Winlink message; there are some fit and finish issues there (images not scaling) that I need to address.

The last bullet point from the migration guide is about error handling of Deferred/Promises. I can watch the Javascript console for unusual error escaping, but beyond that, I don't really have a great strategy for testing that.

xylo04:
I found the issue with the attachment thumbnails, it's discussed in https://jquery.com/upgrade-guide/3.5/. We shouldn't use self-closing HTML tags. Fixed in 3967d08.

martinhpedersen:
What about the v1 -> v2 breaking changes? Anything we might be affected by?

xylo04:
That seems to have been a much more minor change. No API changes, but it dropped support for some old (now ancient) browsers. https://stackoverflow.com/a/23139871/587091

martinhpedersen:
Sounds good 🙂 

Have you tested with real connections over radio?

xylo04:
I hadn't, but I just did, and no issues from Pat.

(It took some doing because I regularly have problems reaching HF stations from my QTH, and setting up AX25 I discovered my attic dual-band antenna had a length of cable go bad. But once the cable was replaced, it worked flawlessly.)

martinhpedersen (approved):

martinhpedersen:
Thanks for taking the time to fix your feed line in order to test this 😄

PR #296, opened 2021-11-09
Update debian control file to require go > 1.16

bseidenberg:
This dependency is required by commit 37f73134 which set a new minimum
version.


xylo04 (approved):

martinhpedersen:
Thanks! 👍

PR #299, opened 2021-11-30
Add support for scriptable message composition

bseidenberg:
Addresses https://github.com/la5nta/pat/issues/

This commit adds support for non-interactive message sending from the command line, in order to support scriptable interactions.
    
To do this, we fork the compose path into 3:
* Interactive (existing)
* Non-Interactive (based on command-line arguments and stdin)
* Reply (synonym for interactive)

_**NOTE:** This is quite literally the first golang code I've written, as well as my first commit to your code base, so please take extra care in the review, and don't spare any feedback - I genuinely want as much as possible to ensure it's correct and idiomatic._

### DIFF
--- a/cli_composer.go
+++ b/cli_composer.go
@@ -102,7 +102,78 @@ func composeMessageHeader(replyMsg *fbb.Message) *fbb.Message {
 	return msg
 }
 
-func composeMessage(replyMsg *fbb.Message) {
+func composeMessage(args []string) {
+	set := pflag.NewFlagSet("compose", pflag.ExitOnError)
+	subject := set.StringP("subject", "s", "", "")
+	attachments := set.StringArrayP("attachment", "a", nil, "")
+	ccs := set.StringArrayP("cc", "c", nil, "")
+	set.Parse(args)
+	recipients := set.Args() // Call after Parse()
+
+	// Check if any args are set. If so, go non-interactive
+	// Otherwise, interactive
+	if (len(*subject) + len(*attachments) + len(*ccs) + len(recipients)) > 0 {
+		noninteractiveComposeMessage(*subject, *attachments, *ccs, recipients)
+	} else {
+		interactiveComposeMessage(nil)
+	}
+
+}
+
+func noninteractiveComposeMessage(subject string, attachments []string,
+	ccs []string, recipients []string) {
+	// We have to verify the args here. Follow the same pattern as main()
+	// We'll allow a missing recipient if CC is present (or vice versa)
+	if len(recipients)+len(ccs) <= 0 {
+		fmt.Fprint(os.Stderr, "ERROR: Missing recipients in non-interactive mode!\n")
+		os.Exit(1)
+	}
+
+	// Subject is optional. Print a mailx style warning
+	if subject == "" {
+		fmt.Fprint(os.Stderr, "Warning: missing subject; hope that's OK\n")
+	}
+
+	msg := fbb.NewMessage(fbb.Private, fOptions.MyCall)
+	msg.SetFrom(fOptions.MyCall)
+	for _, to := range recipients {
+		msg.AddTo(to)
+	}
+	for _, cc := range ccs {
+		msg.AddCc(cc)
+	}
+
+	msg.SetSubject(subject)
+
+	// Handle Attachments. Since we're not interactive, treat errors as fatal so the user can fix
+	for _, filename := range attachments {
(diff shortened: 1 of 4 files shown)


cli_composer.go
```diff
@@ -138,7 +209,7 @@ func composeMessage(replyMsg *fbb.Message) {
 
 	f, err = os.OpenFile(f.Name(), os.O_RDWR, 0o666)
 	if err != nil {
-		log.Fatalf("Unable to read temporary file from editor: %s", err)
+		log.Fatalf("Unable to read temporary file from ed itor: %s", err)
```
xylo04, line 212: This change looks unintentional?

cli_composer.go
```diff
@@ -102,7 +102,78 @@ func composeMessageHeader(replyMsg *fbb.Message) *fbb.Message {
...
+	}
+
+}
+
+func noninteractiveComposeMessage(subject string, attachments []string,
+	ccs []string, recipients []string) {
+	// We have to verify the args here. Follow the same pattern as main()
+	// We'll allow a missing recipient if CC is present (or vice versa)
+	if len(recipients)+len(ccs) <= 0 {
+		fmt.Fprint(os.Stderr, "ERROR: Missing recipients in non-interactive mode!\n")
```
xylo04, line 128: A Golang idiom that took me a while to learn: lower in the call stack you would want to return an error as part of the function return type. However, this is high enough that handling it here is appropriate (as evidenced by the existing code around it). So good job!

xylo04 (changes requested):
Thanks for your first golang PR and first to the repo! Overall this looks good to me, I have only one question/change I'd make, stated below. I'll let Martin make the final pass!

cli_composer.go
bseidenberg: Yeah, I was explicitly following the pattern of the surrounding code (and similar as noted in the comment). Happy to take a look at any resources you have on learning idiomatic handling.

cli_composer.go
bseidenberg: Ooof. Yep.

cli_composer.go
xylo04: I learned that particular one from Martin's code reviews, but this appears to be a good source of idioms: https://go.dev/doc/effective_go#errors

xylo04 (approved):

martinhpedersen:
Sorry I haven't had the time to review this yet. I hope review it soon, so please please bear with me 😊

cli_composer.go
martinhpedersen: > Yeah, I was explicitly following the pattern of the surrounding code (and similar as noted in the comment).

You're refactoring some of the first lines of Go I ever wrote IIRC. So probably not the cleanest and idiomatic code I've written 😅

cli_composer.go
```diff
@@ -102,7 +102,78 @@ func composeMessageHeader(replyMsg *fbb.Message) *fbb.Message {
...
+		if err != nil {
+			fmt.Fprint(os.Stderr, err.Error()+"\nAborting! (Message not posted)\n")
+			os.Exit(1)
+		}
+		msg.AddFile(file)
+	}
+
+	// Read the message body from stdin
+	body, _ := ioutil.ReadAll(os.Stdin)
+	if string(body) == "" {
```
martinhpedersen, line 160: Maybe use `if len(body) == 0 {` instead.

martinhpedersen (approved):
Looks good to me! Nothing much to comment on here, except the points that @xylo04 made in his review 👍 🙂 

If you don't mind, please include another commit to add yourself to the CONTRIBUTORS list in README.md.

Thanks!

cli_composer.go
bseidenberg: I was wondering about that! I even googled for it! Landed at https://stackoverflow.com/questions/18594330/what-is-the-best-way-to-test-for-an-empty-string-in-go which shrugged it off and said both are idiomatic. Happy to change if you prefer.

martinhpedersen: body is of type `[]byte`, not `string`. So that's why I would prefer `len(body) == 0`, instead of casting it to a string just to check if it's length is zero 🙂 

But no big deal! The Go compiler is smart enough not to do an extra allocation such simple `[]byte` to `string` casts.

bseidenberg: Ahhhh, I missed the type difference. Thanks for being explicit. PR sent.

Issue #300, opened 2021-12-01
pat deletes /tmp folder after updating templates


kd2qar:
Rather than just deleting the downloaded file pat attempts to delete the entire /tmp folder on linux. This (hopefully) fails on most systems but I am running pat in a Docker container and noticed that the entire /tmp folder was missing after a while. The problem is in the "forms.go" file where it gets the temp folder and then deletes the entire folder and contents after downloading the file. I suggest it would be better just to delete the downloaded file. The current method could have some problematic issues on a live system if the delete succeeds.

--- forms.go    2021-12-01 07:09:54.146309883 -0500
+++ forms.go    2021-12-01 07:11:38.535008838 -0500
@@ -332,10 +332,11 @@
                filename = fileRe.FindStringSubmatch(dispo)[1]
        }
        dir := os.TempDir()
-       defer os.RemoveAll(dir)
+//     defer os.RemoveAll(dir)
        zipBytes, _ := io.ReadAll(resp.Body)
        defer resp.Body.Close()
        zipFilePath := path.Join(dir, filename)
+       defer os.RemoveAll(zipFilePath)
        err = os.WriteFile(zipFilePath, zipBytes, 0o600)
        if err != nil {
                return fmt.Errorf("can't write update ZIP: %w", err)

The os.RemoveAll() function ( as documented for go) removes both the folder AND the contents.

I have tested the above changes and it solves the issue and cleanly removes the downloaded file only.
-Mark

xylo04:
Hmm, this looks like my bad. I meant to pair this with `os.MkdirTemp` in order to have a temporary subdirectory that I could delete, but I didn't actually do that.

martinhpedersen:
🙂 

I will create a PR for this, which also cleans up a few other things related the download function.

Thanks for reporting this @kd2qar 👍

PR #301, opened 2021-12-01
Improvements to the Forms download functionality

martinhpedersen:
In addition to resolving #300, this also cleans up and improves error handling of the unzip function.

### DIFF
--- a/internal/forms/forms.go
+++ b/internal/forms/forms.go
@@ -15,6 +15,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


internal/forms/forms.go
```diff
@@ -325,80 +326,68 @@ func (m *Manager) downloadAndUnzipForms(downloadLink string) error {
...
-	_ = os.MkdirAll(dest, 0o755)
-
+func unzip(srcArchivePath, dstRoot string) error {
 	// Closure to address file descriptors issue with all the deferred .Close() methods
-	extractAndWriteFile := func(f *zip.File) error {
-		rc, err := f.Open()
-		if err != nil {
-			return err
+	extractAndWriteFile := func(zf *zip.File) error {
+		if zf.FileInfo().IsDir() {
```
xylo04, line 350: Clever way to get away from branching below. Technically this is no longer a perfect reproduction of the zip file, it won't include empty directories, but that's not really behavior that we need for our case.

xylo04 (approved):

PR #302, opened 2021-12-01
Add `env` command

martinhpedersen:
This adds a new CLI command for printing useful env variables.
    
This should be particularly useful when scriping against Pat. It also serves as a sort of documentation of various environment variables that effects Pat.

### DIFF
--- a/env.go
+++ b/env.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"runtime"
+
+	"github.com/la5nta/pat/internal/buildinfo"
+)
+
+func envHandle(_ []string) {
+	writeEnvAll(os.Stdout)
+}
+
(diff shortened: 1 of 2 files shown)


env.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+	"io"
+	"os"
+	"runtime"
+
+	"github.com/la5nta/pat/internal/buildinfo"
+)
+
+func envHandle(_ []string) {
+	writeEnvAll(os.Stdout)
```
xylo04, line 13: Looks useful enough for now. Do you want to extend it later to take an argument similar to `go env`?

xylo04 (approved):

env.go
martinhpedersen: No immediate plans, but maybe? I have been missing a way to programmatically get various directory paths, so. that's the main motivation for this.

PR #308, opened 2021-12-05
Use new self-hosted API for Forms updates

### LINKED ISSUE #306: Host a small service providing latest Forms version and download URL?
The currently [hardcoded URL](https://github.com/la5nta/pat/blob/ea4cfee299e88ecb584d9f9441aa43da6e383896/internal/forms/forms.go#L37) Pat scrapes to get the latest Forms templates, is bound to fail at some point. (Spoiler: It's a winlink.org blog post 😨 ).

How about moving [this code](https://github.com/la5nta/pat/blob/ea4cfee299e88ecb584d9f9441aa43da6e383896/internal/forms/forms.go#L295-L320) to a small cloud-hosted service and expose a JSON API that Pat instances can poll to get this information? Then we could easily update the service on short notice when the current approach finally stops working.

Example API:

```
GET /forms/standard-templates/latest

200 - OK
{
    "version": "1.0.171",
    "archive_url": "https://1drv.ms/u/....."
}
```

We could also consider hosting this as a static file/object (on github pages, AWS S3 or similar), and periodically scrape winlink.org and update the object using cron on any machine available to us. Hosting it this way would give us free fault tolerance and a very efficient cache.

Ping @xylo04 @rainerg2000

martinhpedersen:
This uses our self-hosted API for retrieving the latest Forms version info and download URL.

Closes #306

### DIFF
--- a/internal/forms/forms.go
+++ b/internal/forms/forms.go
@@ -35,7 +35,7 @@ import (
 const (
 	fieldValueFalseInXML = "False"
 	txtFileExt           = ".txt"
-	formsVersionInfoURL  = "https://www.winlink.org/content/all_standard_templates_folders_one_zip_self_extracting_winlink_express_ver_12142016"
+	formsVersionInfoURL  = "https://api.getpat.io/v1/forms/standard-templates/latest"
 )
 
 // Manager manages the forms subsystem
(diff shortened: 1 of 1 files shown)


xylo04 (approved):
Much cleaner!

martinhpedersen:
Thanks 😊 And thank you for setting up the API. This will be much easier to maintain 👍

PR #311, opened 2021-12-15
Add --from/-r option for non-interactive compose.

bseidenberg:
Address https://github.com/la5nta/pat/issues/310https://github.com/la5nta/pat/issues/310

### DIFF
--- a/cli_composer.go
+++ b/cli_composer.go
@@ -104,6 +104,8 @@ func composeMessageHeader(replyMsg *fbb.Message) *fbb.Message {
 
 func composeMessage(args []string) {
 	set := pflag.NewFlagSet("compose", pflag.ExitOnError)
+	// From default is --mycall but it can be overriden with -r
+	from := set.StringP("from", "r", fOptions.MyCall, "")
 	subject := set.StringP("subject", "s", "", "")
 	attachments := set.StringArrayP("attachment", "a", nil, "")
 	ccs := set.StringArrayP("cc", "c", nil, "")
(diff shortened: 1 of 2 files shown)


bseidenberg:
Easy enough. Basic smoke testing seems fine - I can impersonate a known shared mailbox just fine. (*Sigh*)

cli_composer.go
```diff
@@ -104,6 +104,7 @@ func composeMessageHeader(replyMsg *fbb.Message) *fbb.Message {
 
 func composeMessage(args []string) {
 	set := pflag.NewFlagSet("compose", pflag.ExitOnError)
+	from := set.StringP("from", "r", "", "")
 	subject := set.StringP("subject", "s", "", "")
```
martinhpedersen, line 108: How about setting the default value to fOptions.Mycall here? It would simplify the if / else block below 🙂

cli_composer.go
```diff
@@ -143,7 +144,12 @@ func noninteractiveComposeMessage(subject string, attachments []string,
 	}
 
 	msg := fbb.NewMessage(fbb.Private, fOptions.MyCall)
-	msg.SetFrom(fOptions.MyCall)
+	if len(from) > 0 {
+		msg.SetFrom(from)
+
```
martinhpedersen, line 149: I think we should refer to this as "sender address/callsign" to highlight that this could be used to set an email address as the sender as well as a callsign.

Also, the correct term is "tactical addresses" IIRC.

martinhpedersen: Sorry abou that. This comment is on the wrong line. Should be seen with regards to the usage text.

cli_composer.go
bseidenberg: I can do that. I figured the code is more explicit this way, but happy to go that route. (I also looked for the ternary operator, but that's not a golang thing).

cli_composer.go
bseidenberg: Done (though I wouldn't want to encourage a non-WL email address for the reasons you mentioned previously)

I went with "Address to send from. Default is your call from config or --mycall, but can be specified to use tactical addresses."

martinhpedersen (approved):
Perfect 🥇 Thank you 😄

PR #312, opened 2021-12-29
bump harenber/ptc-go to v2.2.3 (to prepare Android support)

harenber:
ptc-go v2.2.3 has an updated dependency to albenik/go-serial which brings Android support (see https://github.com/albenik/go-serial/issues/29). Furthermore, on every connect using a PACTOR modem, the code syncs now date and time on the modem.


xylo04 (approved):

martinhpedersen:
Thanks 😊

PR #323, opened 2022-02-16
Graceful shutdown through context cancellation

### LINKED ISSUE #314: Can we handle interrupts in listen.go instead of exhange.go?
Hi @martinhpedersen ,

for the PACTOR listen mode, I have to set the modem into the WA8DED hostmode as soon as Pat is started with "-l pactor". However, when the user ends Pat with CTRL-C, the pactor driver is only notified when a current exchange is going on (signal handler in exchange.go), not when no connection was active. So there is no way to set the modem back to terminal mode. While this is not a problem for Pat (the PACTOR driver can handle this), it may be confusing for the user.

edit: probably listen.go is also not a good place, as we need the interrupt when the user aborts a connect as well. Is there a upper method than would be suitable?

Cheers

  Torsten

martinhpedersen:
By doing graceful shutdown, we're able to properly close all of the open resources such as modem connections and ongoing sessions.

This will allow transport drivers controlling stateful TNCs (e.g. ptc-go) to recover the initial TNC state before termination.

Resolves #314.

------

This approach uses Go's [context package](https://pkg.go.dev/context). Here are some posts explaining the concept:
- http://p.agnihotry.com/post/understanding_the_context_package_in_golang/
- https://go.dev/blog/context

### DIFF
--- a/cli_composer.go
+++ b/cli_composer.go
@@ -8,6 +8,7 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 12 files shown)


xylo04 (approved):
A lot of plumbing, but it's worth it!

martinhpedersen:
Hehe, indeed! Thanks for the review 🙂

Issue #328, opened 2022-03-13
Some "insertion tags" are missing (like GPS)


n2ygk:
I noticed that my gpsd-connected GPS position was not being automatically populated in the GENERAL Forms/Winlink Check In form. It appears that not all the [insertion tags](https://www.winlink.org/sites/default/files/RMSE_FORMS/insertion_tags.zip) are implemented in forms.go:

https://github.com/la5nta/pat/blob/ccb493474debb2a84ee2ab60b292327426c86f8a/internal/forms/forms.go#L737-L755

Here's an example from the browser console showing no replacement of the value of  `{GPS_SIGNED_DECIMAL}` vs. `{MsgSender}` that is replaced by the code above:
```
> ExtGPS
<input name=​"ExtGPS" type=​"hidden" id=​"ExtGPS" value=​"{GPS_SIGNED_DECIMAL}​">​
> MsgSender
<input title=​"Call sign or tactical being used to send mail" readonly=​"readonly" maxlength=​"16" size=​"17" id=​"MsgSender" name=​"MsgSender" value=​"N2YGK">​
```

These correspond to the following lines in the input HTML form:
```
winlink$ grep  -n {GPS_SIG ~/Library/Application\ Support/pat/Standard_Forms/GENERAL\ Forms/Winlink_Check_In_Initial.html
506: if (ans  != '{GPS_SIGNED_DECIMAL}') {document.getElementById('GPS2').value = ans};
1494:<input name="ExtGPS" type="hidden" id="ExtGPS" value="{GPS_SIGNED_DECIMAL}" />
1501:<input name="GPS" type="hidden" id="GPS" value="{GPS_SIGNED_DECIMAL}" />
1513:<input name="GPS" type="hidden" id="GPS" value="{GPS_SIGNED_DECIMAL}" />
winlink$ 

$ grep -n {MsgSender} ~/Library/Application\ Support/pat/Standard_Forms/GENERAL\ Forms/Winlink_Check_In_Initial.html 
1477:    <input title="Call sign or tactical being used to send mail" readonly="readonly" maxlength="16" size="17" id="MsgSender"  name="MsgSender" value="{MsgSender}" />
1573:<input name="testCall" type="hidden" id="testCall" value="{MsgSender}" size="14" maxlength="</table>13" />
```

martinhpedersen:
Closing this one as duplicate of #338.

PR #330, opened 2022-03-13
Implement missing form template GPS-related insertion tags

n2ygk:
Fixes: #328 

Implements several missing insertion tags for form templates (some of which are not documented other than by reading the HTML):

- {GPS} -  degrees and decimal minutes with Northing and Easting: 46-22.77N 121-35.01W
- {GPS_DECIMAL} - decimal degrees with Northing and Easting: 46.3795N 121.5835W
- {GPS_SIGNED_DECIMAL} - signed decimal degrees: 41.1234 -73.4567
- {Latitude} and {latitude} - signed decimal latitude degrees: 41.1234
- {Longitude} and {longitude} - signed decimal longitude degrees: -73.4567
- {GridSquare} - maidenhead grid square locator: FN31cd
- {GPSValid} - "YES " or "NO " to indicate GPS position is valid

Some of these values are substituted in forms but not used (like GridSquare).

### DIFF
--- a/internal/forms/forms.go
+++ b/internal/forms/forms.go
@@ -18,6 +18,7 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
+	"math"
 	"net/http"
 	"os"
 	"path"
@@ -31,6 +32,10 @@ import (
 	"unicode/utf8"
 
 	"github.com/dimchansky/utfbom"
+	"github.com/la5nta/pat/cfg"
+	"github.com/la5nta/pat/internal/debug"
+	"github.com/la5nta/pat/internal/gpsd"
+	"github.com/pd0mz/go-maidenhead"
 )
 
 const (
@@ -60,6 +65,7 @@ type Config struct {
 	AppVersion string
 	LineReader func() string
 	UserAgent  string
+	GPSd       cfg.GPSdConfig
 }
 
 // Form holds information about a Winlink form template
@@ -705,6 +711,99 @@ func (m *Manager) findAbsPathForTemplatePath(tmplPath string) (string, error) {
 	return retVal, nil
 }
 
+// gpsPos returns the current GPS Position
+func (m *Manager) gpsPos() (gpsd.Position, error) {
+	addr := m.config.GPSd.Addr
+	if addr == "" {
+		return gpsd.Position{}, errors.New("GPSd: not configured.")
+	}
+	if !m.config.GPSd.AllowForms {
+		return gpsd.Position{}, errors.New("GPSd: allow_forms is disabled. GPS position will not be available in form templates.")
+	}
+
+	conn, err := gpsd.Dial(addr)
+	if err != nil {
+		log.Printf("GPSd daemon: %s", err)
+		return gpsd.Position{}, err
+	}
+
+	defer conn.Close()
+
+	conn.Watch(true)
+
+	log.Println("Waiting for position from GPSd...")
+	// TODO: make the GPSd timeout configurable
+	return conn.NextPosTimeout(3 * time.Second)
+}
+
+type gpsStyle int
+
+const (
+	// documentation: https://www.winlink.org/sites/default/files/RMSE_FORMS/insertion_tags.zip
+	signedDecimal gpsStyle = iota // 41.1234 -73.4567
+	decimal                       // 46.3795N 121.5835W
+	degreeMinute                  // 46-22.77N 121-35.01W
+)
+
+func gpsFmt(style gpsStyle, pos gpsd.Position) string {
+	var (
+		northing   string
+		easting    string
+		latDegrees int
+		latMinutes float64
+		lonDegrees int
+		lonMinutes float64
+	)
+
+	noPos := gpsd.Position{}
+	if pos == noPos {
+		return "(Not available)"
+	}
+	switch style {
+	case degreeMinute:
+		{
+			latDegrees = int(math.Trunc(math.Abs(pos.Lat)))
+			latMinutes = (math.Abs(pos.Lat) - float64(latDegrees)) * 60
+			lonDegrees = int(math.Trunc(math.Abs(pos.Lon)))
+			lonMinutes = (math.Abs(pos.Lon) - float64(lonDegrees)) * 60
+		}
+		fallthrough
+	case decimal:
+		{
+			if pos.Lat >= 0 {
+				northing = "N"
+			} else {
+				northing = "S"
+			}
+			if pos.Lon >= 0 {
+				easting = "E"
+			} else {
+				easting = "W"
+			}
+		}
+	}
+
+	switch style {
+	case signedDecimal:
+		return fmt.Sprintf("%.4f %.4f", pos.Lat, pos.Lon)
+	case decimal:
+		return fmt.Sprintf("%.4f%s %.4f%s", math.Abs(pos.Lat), northing, math.Abs(pos.Lon), easting)
+	case degreeMinute:
+		return fmt.Sprintf("%02d-%05.2f%s %03d-%05.2f%s", latDegrees, latMinutes, northing, lonDegrees, lonMinutes, easting)
+	default:
+		return "(Not available)"
+	}
+}
+
+func posToGridSquare(pos gpsd.Position) string {
+	point := maidenhead.NewPoint(pos.Lat, pos.Lon)
+	gridsquare, err := point.GridSquare()
+	if err != nil {
+		return ""
+	}
+	return gridsquare
+}
+
 func (m *Manager) fillFormTemplate(absPathTemplate string, formDestURL string, placeholderRegEx *regexp.Regexp, formVars map[string]string) (string, error) {
 	fUnsanitized, err := os.Open(absPathTemplate)
 	if err != nil {
@@ -733,6 +832,15 @@ func (m *Manager) fillFormTemplate(absPathTemplate string, formDestURL string, p
 	nowDateUTC := now.UTC().Format("2006-01-02Z")
 	nowTimeUTC := now.UTC().Format("15:04:05Z")
 	udtg := strings.ToUpper(now.UTC().Format("021504Z Jan 2006"))
+	nowPos, err := m.gpsPos()
+	var validPos string
+	if err != nil {
+		validPos = "NO"
+		debug.Printf(fmt.Sprint(err))
+	} else {
+		validPos = "YES"
+		debug.Printf("GPSd position: %s", gpsFmt(signedDecimal, nowPos))
+	}
 
 	scanner := bufio.NewScanner(bytes.NewReader(sanitizedFileContent))
 	for scanner.Scan() {
@@ -750,6 +858,17 @@ func (m *Manager) fillFormTemplate(absPathTemplate string, formDestURL string, p
 		l = strings.ReplaceAll(l, "{UDTG}", udtg)
 		l = strings.ReplaceAll(l, "{Time}", nowTime)
 		l = strings.ReplaceAll(l, "{UTime}", nowTimeUTC)
+		l = strings.ReplaceAll(l, "{GPS}", gpsFmt(degreeMinute, nowPos))
+		l = strings.ReplaceAll(l, "{GPS_DECIMAL}", gpsFmt(decimal, nowPos))
+		l = strings.ReplaceAll(l, "{GPS_SIGNED_DECIMAL}", gpsFmt(signedDecimal, nowPos))
+		// Lots of undocumented tags found in the Winlink check in form.
+		// Note also various ways of capitalizing. Perhaps best to do case insenstive string replacements....
+		l = strings.ReplaceAll(l, "{Latitude}", fmt.Sprintf("%.4f", nowPos.Lat))
+		l = strings.ReplaceAll(l, "{latitude}", fmt.Sprintf("%.4f", nowPos.Lat))
+		l = strings.ReplaceAll(l, "{Longitude}", fmt.Sprintf("%.4f", nowPos.Lon))
+		l = strings.ReplaceAll(l, "{longitude}", fmt.Sprintf("%.4f", nowPos.Lon))
+		l = strings.ReplaceAll(l, "{GridSquare}", posToGridSquare(nowPos))
+		l = strings.ReplaceAll(l, "{GPSValid}", fmt.Sprintf("%s ", validPos))
 		if placeholderRegEx != nil {
 			l = fillPlaceholders(l, placeholderRegEx, formVars)
 		}
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Copyright (c) 2020 Martin Hebnes Pedersen LA5NTA
 * LA3QMA - Kai Günter Brandt
 * LA4TTA - Erlend Grimseid
 * LA5NTA - Martin Hebnes Pedersen
+* N2YGK - Alan Crosswell
 * VE7GNU - Doug Collinge
 * W6IPA  - JC Martin
 * WY2K - Benjamin Seidenberg
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -226,11 +226,15 @@ type GPSdConfig struct {
 	// Caution: Your GPS position will be accessible to any network device able to access Pat's HTTP interface.
 	EnableHTTP bool `json:"enable_http"`
 
-	// Use server time instead of timestamp provided by GPSd (e.g for older GPS
-	// device with week roll-over issue)
+	// Allow Winlink forms to use GPSd for aquiring your position.
+	//
+	// Caution: Your current GPS position will be automatically injected, without your explicit consent, into forms requesting such information.
+	AllowForms bool `json:"allow_forms"`
+
+	// Use server time instead of timestamp provided by GPSd (e.g for older GPS device with week roll-over issue).
 	UseServerTime bool `json:"use_server_time"`
 
-	// Address and port of GPSd server (e.g. localhost:2947)
+	// Address and port of GPSd server (e.g. localhost:2947).
 	Addr string `json:"addr"`
 }
 
@@ -272,6 +276,7 @@ var DefaultConfig = Config{
 	},
 	GPSd: GPSdConfig{
 		EnableHTTP:    false, // Default to false to help protect privacy of unknowing users (see github.com//issues/146)
+		AllowForms:    false, // Default to false to help protect location privacy of unknowing users
 		UseServerTime: false,
 		Addr:          "localhost:2947", // Default listen address for GPSd
 	},
--- a/main.go
+++ b/main.go
@@ -346,6 +346,7 @@ func main() {
 		AppVersion: buildinfo.VersionStringShort(),
 		UserAgent:  buildinfo.UserAgent(),
 		LineReader: readLine,
+		GPSd:       config.GPSd,
 	})
 
 	// Make sure we clean up on exit, closing any open resources etc.


n2ygk:
@xylo04 I thought I'd give it a try.

n2ygk:
(ignore those extra commits -- I was having issues pushing from my virtualbox...)

internal/forms/forms.go
```diff
@@ -704,6 +705,49 @@ func (m *Manager) findAbsPathForTemplatePath(tmplPath string) (string, error) {
...
+	// <GPS> - Inserts the current GPS location (if available)
+	// Ex: 46-22.77N 121-35.01W
+	// <GPS_DECIMAL> - Inserts the current GPS location (if available)
+	// Ex: 46-22.77N 121-35.01W (this example is probably incorrect)
+	// Ex: 46.3795N 121.5835W
+	// <GPS_SIGNED_DECIMAL> - Inserts the current GPS location as a signed decimal latitude/longitude (if available)
+	// Ex: 46.3795 -121.5835
+
+
+	addr := "localhost:2947" // TODO: get from config
```
xylo04, line 724: Within the Manager struct (most of the functions here are actually methods on that struct), the config object is available. So as long as that's in scope, getting this value should be as simple as m.config.GPSd.Addr.

internal/forms/forms.go
```diff
@@ -732,6 +776,9 @@ func (m *Manager) fillFormTemplate(absPathTemplate string, formDestURL string, p
 	nowDateUTC := now.UTC().Format("2006-01-02Z")
 	nowTimeUTC := now.UTC().Format("15:04:05Z")
 	udtg := strings.ToUpper(now.UTC().Format("021504Z Jan 2006"))
+	nowGPS := gpsPos(false, false)
```
xylo04, line 779: I would suggest separating the tasks of getting the position from the GPS device vs formatting the position into different methods. Since we're having to do some costly communication with the GPS device, doing that three times seems like overkill.

xylo04 (changes requested):
@n2ygk this looks like a great first crack! Thanks for putting this together. Does it work for you locally? If so, that's super exciting!

internal/forms/forms.go
n2ygk: Actually I tried this at first and there's a local version of Config that appears not to have GPSd.... 

https://github.com/la5nta/pat/blob/ccb493474debb2a84ee2ab60b292327426c86f8a/internal/forms/forms.go#L55-L62

n2ygk:
@xylo04 This is working properly for the GPS form but not the Winlink Check In form. Not sure why just yet other than suspecting I'm either not formatting the signed decimal string correctly or the form is buggy. The LAT gets filled in properly but the LON doesn't.

See attached before and after screen shots for the GPS Position report form. In before, the tag has not been substituted and still showed as `{GPS}` while in the after, it has my position.
Before:
before

After:
after

Clicking on "Click to use connected GPS LAT LON" does the right thing and fills in all the other fields on the form.

n2ygk:
I got the Winlink Check In form working by reverse-engineering and finding several undocumented insertion tags which I've added in [32e7106](https://github.com/la5nta/pat/pull/330/commits/32e71069c05954e2003caa5442e5038ef9fd8130).

There's a lot of random capitalization going on and some tags that are both supplied by the HTTP server and calculated in the form -- notably Grid Square.

internal/forms/forms.go
xylo04: Ah, I see you're right. In that case you'll need to add a new field to forms.Config to hold the GPSd address, and plumb it though in main.go where we call forms.NewManager.

internal/forms/forms.go
```diff
@@ -749,6 +835,18 @@ func (m *Manager) fillFormTemplate(absPathTemplate string, formDestURL string, p
 		l = strings.ReplaceAll(l, "{UDTG}", udtg)
 		l = strings.ReplaceAll(l, "{Time}", nowTime)
 		l = strings.ReplaceAll(l, "{UTime}", nowTimeUTC)
+		l = strings.ReplaceAll(l, "{GPS}", gpsFmt(degreeMinute, nowPos))
+		l = strings.ReplaceAll(l, "{GPS_DECIMAL}", gpsFmt(decimal, nowPos))
+		l = strings.ReplaceAll(l, "{GPS_SIGNED_DECIMAL}", gpsFmt(signedDecimal, nowPos))
+		// Lots of undocumented tags found in the Winlink check in form.
```
xylo04, line 841: Good find, but... yeesh.

xylo04 (changes requested):
Nice improvements! Getting closer.

martinhpedersen:
Thanks for working on this @n2ygk 🙂 

It looks like you haven't configured your editor to run `go fmt`? If you haven't, please remember `go fmt` before squashing and finalizing this PR. Thanks!

PS: We have a [CONTRIBUTING.md](https://github.com/la5nta/pat/blob/master/CONTRIBUTING.md) with some guidance to first time contributors. It's not the best, but worth the read hopefully 🙂

internal/forms/forms.go
```diff
@@ -704,6 +706,82 @@ func (m *Manager) findAbsPathForTemplatePath(tmplPath string) (string, error) {
...
+	} else {
+		return gpsd.Position{}, errors.New("GPSd: not configured.")
+	}
+}
+
+type gpsStyle int
+
+const (
+	// documentation: https://www.winlink.org/sites/default/files/RMSE_FORMS/insertion_tags.zip
+        signedDecimal gpsStyle = iota // 41.1234 -73.4567
```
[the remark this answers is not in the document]
n2ygk: I googled "golang equivalent of enum"

internal/forms/forms.go
n2ygk: Yeah, ugh.

n2ygk:
> Thanks for working on this @n2ygk 🙂
> 
> It looks like you haven't configured your editor to run `go fmt`? If you haven't, please remember `go fmt` before squashing and finalizing this PR. Thanks!
> 
> PS: We have a [CONTRIBUTING.md](https://github.com/la5nta/pat/blob/master/CONTRIBUTING.md) with some guidance to first time contributors. It's not the best, but worth the read hopefully 🙂

@martinhpedersen: no sorry I didn't run `go fmt` but when I do the changes are to unrelated files:
```
vagrant@debian11:~/src/pat$ go fmt
http_multipart.go
riglist.go
```
Am I missing something? I can certainly rebase. Will do that now. Thanks for the reminder.

martinhpedersen:
> ... but when I do the changes are to unrelated files:
> 
> ```
> vagrant@debian11:~/src/pat$ go fmt
> http_multipart.go
> riglist.go
> ```
> Am I missing something?

These seems to be related to a change in how build tags are handled (>Go 1.17). I will fix those 👍 

Running `go fmt` will only modify files in the current directory. You could use `go fmt ./...` to format sub-packages as well, but most Go developers configure their editor to run `go fmt` on the target file on save. Most Go-enabled IDEs will do this for you out of the box, and plain old editors like `vim` and `emacs` have various plugins to support this. (I'm a vim user myself, and use [vim-go](https://github.com/fatih/vim-go)).

n2ygk:
@martinhpedersen Ah. Thanks for the pointer. My editor is Emacs and my go knowledge is zero.
I found a few others besides the one I touched. Looks like the same build tags issue.  I'll fix forms.go and push.
```
vagrant@debian11:~/src/pat$ go fmt ./...
http_multipart.go
riglist.go
internal/forms/forms.go
internal/osutil/rlimit_unix.go
internal/osutil/rlimit_freebsd.go
internal/osutil/rlimit_windows.go
```

n2ygk:
@martinhpedersen fixed in my .emacs: `(add-hook 'before-save-hook #'gofmt-before-save)"`

martinhpedersen:
> fixed in my .emacs: `(add-hook 'before-save-hook #'gofmt-before-save)"`

Nice! 🙂 👍 

One more thing! We generally merge PRs to `develop` to keep `master` on track with the latest release. If you don't mind, please update the base branch of this PR and rebase on top of `develop`. This is one of the things missing in CONTRIBUTORS.md (note to self). Sorry about that.

Thanks!

martinhpedersen:
@n2ygk - Are you still working on this, or waiting for a new review and/or merge?

n2ygk:
> @n2ygk - Are you still working on this, or waiting for a new review and/or merge?

I'm done. Waiting for review/merge.

internal/forms/forms.go
```diff
@@ -705,6 +709,108 @@ func (m *Manager) findAbsPathForTemplatePath(tmplPath string) (string, error) {
...
+	degreeMinute                  // 46-22.77N 121-35.01W
+)
+
+func gpsFmt(style gpsStyle, pos gpsd.Position) string {
+	var northing string
+	var easting string
+	var latDegrees int
+	var latMinutes float64
+	var lonDegrees int
+	var lonMinutes float64
```
martinhpedersen, line 753: It's preferable to define variables as local as possible. I think this is a symptom of the complexity introduced by accepting the style parameter instead of having multiple implementing functions for the different styles.

PS: When declaring lots of variables like this, it is a bit cleaner to use this syntax:

```go
var (
   northing string
   easting string
   latDegrees int
   latMinutes float64
   lonDegrees int
   lonMinutes float64
)
```

internal/forms/forms.go
```diff
@@ -705,6 +709,108 @@ func (m *Manager) findAbsPathForTemplatePath(tmplPath string) (string, error) {
...
+	case decimal:
+		return fmt.Sprintf("%.4f%s %.4f%s", math.Abs(pos.Lat), northing, math.Abs(pos.Lon), easting)
+	case degreeMinute:
+		return fmt.Sprintf("%02d-%05.2f%s %03d-%05.2f%s", latDegrees, latMinutes, northing, lonDegrees, lonMinutes, easting)
+	default:
+		return "(Not available)"
+	}
+}
+
+func posToGridSquare(pos gpsd.Position) string {
```
martinhpedersen, line 795: We're already depending on github.com/pd0mz/go-maidenhead for doing maidenhead calculation. Maybe it provides some of this functionality already?

internal/forms/forms.go
```diff
@@ -750,6 +863,18 @@ func (m *Manager) fillFormTemplate(absPathTemplate string, formDestURL string, p
...
+		l = strings.ReplaceAll(l, "{GPS_SIGNED_DECIMAL}", gpsFmt(signedDecimal, nowPos))
+		// Lots of undocumented tags found in the Winlink check in form.
+		// Note also various ways of capitalizing. Perhaps best to do case insenstive string replacements....
+		l = strings.ReplaceAll(l, "{Latitude}", fmt.Sprintf("%.4f", nowPos.Lat))
+		l = strings.ReplaceAll(l, "{latitude}", fmt.Sprintf("%.4f", nowPos.Lat))
+		l = strings.ReplaceAll(l, "{Longitude}", fmt.Sprintf("%.4f", nowPos.Lon))
+		l = strings.ReplaceAll(l, "{longitude}", fmt.Sprintf("%.4f", nowPos.Lon))
+		l = strings.ReplaceAll(l, "{GridSquare}", posToGridSquare(nowPos))
+		l = strings.ReplaceAll(l, "{GPSValid}", fmt.Sprintf("%s ", validPos))
+		// l = strings.ReplaceAll(l, "{InternetAvailable}", "TO DO")
```
martinhpedersen, line 877: TODO?

martinhpedersen (changes requested):
Thanks for working on this @n2ygk 😄 

I think we're almost ready to merge this. Please see my inline comments.

Thank you!

internal/forms/forms.go
```diff
@@ -705,6 +709,108 @@ func (m *Manager) findAbsPathForTemplatePath(tmplPath string) (string, error) {
...
 }
 
+func (m *Manager) gpsPos() (gpsd.Position, error) {
+	// return current GPS Position
+
+	addr := m.config.GPSd.Addr
+	if addr != "" {
+		conn, err := gpsd.Dial(addr)
+		if err != nil {
+			log.Printf("GPSd daemon: %s", err)
```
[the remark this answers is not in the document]
n2ygk: yep

internal/forms/forms.go
```diff
@@ -705,6 +709,108 @@ func (m *Manager) findAbsPathForTemplatePath(tmplPath string) (string, error) {
...
+		defer conn.Close()
+
+		conn.Watch(true)
+
+		log.Println("Waiting for position from GPSd...")
+		pos, err := conn.NextPosTimeout(3 * time.Second) // TODO: configurable
+		if err != nil {
+			log.Printf("GPSd: %s", err)
+		} else {
+			log.Printf("GPSd: %f %f", pos.Lat, pos.Lon)
```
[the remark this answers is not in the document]
n2ygk: I kinda like seeing it in the console log... but sure, I can remove it.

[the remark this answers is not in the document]
n2ygk: Also, it definitively ends the "Waiting for position from GPSd ..." which can take up to 3 seconds before either succeeding or failing.

internal/forms/forms.go
```diff
@@ -705,6 +709,108 @@ func (m *Manager) findAbsPathForTemplatePath(tmplPath string) (string, error) {
...
+type gpsStyle int
+
+const (
+	// documentation: https://www.winlink.org/sites/default/files/RMSE_FORMS/insertion_tags.zip
+	signedDecimal gpsStyle = iota // 41.1234 -73.4567
+	decimal                       // 46.3795N 121.5835W
+	degreeMinute                  // 46-22.77N 121-35.01W
+)
+
+func gpsFmt(style gpsStyle, pos gpsd.Position) string {
```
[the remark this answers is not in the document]
n2ygk: I thought I was being so fancy google what the go version of a C enum was... ;-)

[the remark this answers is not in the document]
n2ygk: Some of the paths reuse code, so...

internal/forms/forms.go
n2ygk: I went through the forms and found a bunch of undocumented insertion tags. One if them was `InternetAvailable` which presumably somehow determines if Internet connectivity is currently in place. I didn't know how to determine that, and it wasn't really relevant to getting the GPSd stuff working but I wanted to document that there is this (and perhaps other) insertion tags that still need to be implemented. I will remove this line and create an Issue instead.

n2ygk: #338

internal/forms/forms.go
n2ygk: Well, yes, there is! Thanks for that.

internal/forms/forms.go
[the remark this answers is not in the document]
n2ygk: @martinhpedersen I've left this one as is. Let me know if that's OK.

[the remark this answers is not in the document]
martinhpedersen: > I thought I was being so fancy googling what the go version of a C enum was... ;-)

Oh yes, your `iota` usage was spot on! 👍 

Sure! Let's keep it this way. No problem 🙂

internal/forms/forms.go
```diff
@@ -709,30 +710,28 @@ func (m *Manager) findAbsPathForTemplatePath(tmplPath string) (string, error) {
...
-		pos, err := conn.NextPosTimeout(3 * time.Second) // TODO: configurable
-		if err != nil {
-			log.Printf("GPSd: %s", err)
-		} else {
-			log.Printf("GPSd: %f %f", pos.Lat, pos.Lon)
-		}
-		return pos, err
-	} else {
-		return gpsd.Position{}, errors.New("GPSd: not configured.")
+	defer conn.Close()
```
martinhpedersen, line 722: Very dangerous to defer conn.Close before checking that error. If there is an error, changes are `conn` is nil and you will get a nil pointer panic once the deferred Close call is executed 😅

internal/forms/forms.go
```diff
@@ -709,30 +710,28 @@ func (m *Manager) findAbsPathForTemplatePath(tmplPath string) (string, error) {
...
+		log.Printf("GPSd daemon: %s", err)
+		return gpsd.Position{}, err
 	}
+
+	conn.Watch(true)
+
+	log.Println("Waiting for position from GPSd...")
+	pos, err := conn.NextPosTimeout(3 * time.Second) // TODO: configurable
+
+	return pos, err
```
martinhpedersen, line 734:
```go
return conn.NextPosTimeout(3 * time.Second)
```

That would be even easier to read 🙂

internal/forms/forms.go
```diff
@@ -793,22 +794,12 @@ func gpsFmt(style gpsStyle, pos gpsd.Position) string {
...
-	adjLon := pos.Lon + 180.0
-	grid = grid + string(U[int(adjLon/20.0)]) + string(U[int(adjLat/10.0)])
-	nLat := int(adjLat) % 10
-	nLon := int(adjLon/2) % 10
-	grid = grid + fmt.Sprintf("%d%d", nLon, nLat)
-	rLat := (adjLat - math.Trunc(adjLat)) * 60.0
-	rLon := (adjLon - 2*math.Trunc(adjLon/2)) * 60.0
-	grid = grid + string(L[int(math.Trunc(rLon/5))]) + string(L[int(math.Trunc(rLat/2.5))])
-	return grid
+	point := maidenhead.NewPoint(pos.Lat, pos.Lon)
```
martinhpedersen, line 797: Nice! 🎉

martinhpedersen (changes requested):
Almost there!

n2ygk:
I supposed I should rebase this.... Standby.

martinhpedersen (commented):
Since our default config provides a [default gpsd address](https://github.com/la5nta/pat/blob/ccb493474debb2a84ee2ab60b292327426c86f8a/cfg/config.go#L300) this will always throw an error for users without GPSd.

I think we need a more graceful handling of this condition than printing the standard tcp dial error each time those users fills in a form template.

Maybe we should simply omit the error if it's a network dial error?

n2ygk:
> Since our default config provides a [default gpsd address](https://github.com/la5nta/pat/blob/ccb493474debb2a84ee2ab60b292327426c86f8a/cfg/config.go#L300) this will always throw an error for users without GPSd.
> 
> I think we need a more graceful handling of this condition than printing the standard tcp dial error each time those users fills in a form template.
> 
> Maybe we should simply omit the error if it's a network dial error?

How about changing the default config to not list a gpsd address? It's kind of wasteful to attempt those dials irrespective of whether errors are being logged. I wouldn't want to not log errors if there is in fact expected to be a gpsd running.

martinhpedersen:
I think we should avoid comprimising a good default value in our config just to keep from logging errors for users of these forms that don't care about using GPSd. The default config is very convenient for the other GPSd related features, requiring no config on most systems.

How about adding a `enable_forms` config option similar to `enable_http`? That would be a good for security as well, so we don't automatically expose a user's position without their concent.

https://github.com/la5nta/pat/blob/587580f4bce65b742bb01052ee082d3388ce0be0/cfg/config.go#L227

martinhpedersen:
@n2ygk - I've added the additional gpsd config option in 07b56654aea82754cae0f506b070b1e2787ab990. You can either cherry-pick that and glue it together in this PR, or let me know and I'll fix the rest 🙂 

Also if you could add yourself to the Contributors list if READM.md, that would be great. Thanks!

n2ygk:
@martinhpedersen I need a little help here. I've made the change but when trying to compile and run there appears to have been some change to the web js code that throws this error: `Uncaught ReferenceError: onFormLaunching is not defined` I guess I'm looking for the equivalent of `make install` that installed not only the pat binary but the web files....

martinhpedersen:
@n2ygk - I think you should rebase again, and solve that conflict in README.md.

I don't fully understand why you would get such an error. Probably not related to this PR anyway, but I will check it out once you've rebased 🙂 

Thanks!!

n2ygk:
> @n2ygk - I think you should rebase again, and solve that conflict in README.md.
> 
> I don't fully understand why you would get such an error. Probably not related to this PR anyway, but I will check it out once you've rebased 🙂
> 
> Thanks!!

I suspect this may be related to 7850c5f036cab4892429352a9318aaa685722d94 which I just found now that I've rebased;-)

n2ygk:
@martinhpedersen OK so this is looking like it does the right thing now. It's being very quiet about the error message per your prior recommendation so the user will not realize that the GPSd stuff needs to add `"allow_forms": true`. I suppose a wiki page update is warranted as well. Let me know if you'd like me to submit that.

internal/forms/forms.go
```diff
@@ -705,6 +710,99 @@ func (m *Manager) findAbsPathForTemplatePath(tmplPath string) (string, error) {
 	return retVal, nil
 }
 
+// gpsPos returns the current GPS Position
+func (m *Manager) gpsPos() (gpsd.Position, error) {
```
martinhpedersen, line 714: Since the error returned from this function is never used for anything but to check for success, I think we should replace this with a "ok" boolean. Then you could omit the check for "noPos" as well 🙂 

All these error messages could be useful for debugging, so I suggest logging them instead using our debug logger [debug.Printf](https://pkg.go.dev/github.com/la5nta/pat/internal/debug#Printf).

```
func (m *Manager) gpsPos() (pos gpsd.Position, ok bool) {

}
```

I'd be happy to do these last minute tweaks after merging if you prefer. Just say the word 🙂

martinhpedersen (commented):
Once last thing... (see inline comment).

martinhpedersen:
@n2ygk 
> I suppose a wiki page update is warranted as well. Let me know if you'd like me to submit that.

I agree! If you don't mind, yes please 😄

internal/forms/forms.go
n2ygk: The code currently propagates the return from NextPosTimeout() so this will become quite a bit uglier as would the NoPos checking since gpsFmt is called with a Position.

I've added a couple of debug Printf of the gpsPos error.

Feel free to make the tweaks you suggest if you disagree.

martinhpedersen: Sure, I see what you mean 🙂 

As a general rule of thumb I think it's better to *either* log or return an error, not both. It helps with readability and makes the responsibilities clearer IMO.

I think we should merge this now, and I'll see what can be done with regards to this afterwards.

Thank you 😄

martinhpedersen (approved):

martinhpedersen:
Thanks for working on this @n2ygk! Highly appreciated 🙂

n2ygk:
> @n2ygk
> 
> > I suppose a wiki page update is warranted as well. Let me know if you'd like me to submit that.
> 
> I agree! If you don't mind, yes please 😄

@martinhpedersen done: https://github.com/la5nta/pat/wiki/The-web-GUI

olewsaa:
I've put together a Python script that request GPS position from the Signal K server. It respond with nn.dddd nn.dddd, but I included a routine that convert the output to a grid reference.  It would be benefical if Signal K could be added as a source. File is [get.grid.py](https://github.com/olewsaa/Yacht-computer/blob/master/pat/get.grid.py). Not all RPis have a local GPS attached.

PR #335, opened 2022-03-21
Add --p2p-only option to compose

### LINKED ISSUE #334: Add --p2p-only to pat compose
> Is it possible to set the P2P flag when using compose from the command line? Not interactive mode, but something that could be used in a script.
> 
> 73, de @km4ack

martinhpedersen:
Resolves #334


xylo04 (approved):
LGTM

PR #341, opened 2022-04-03
Fix loading forms on `develop` branch

xylo04:
After af07b47b05e2af72d9de1b72c709f6c4a94a8d88, trying to launch a form from the compose dialog is broken. That's because the javascript functions are no longer in the global namespace, but clicking a form name was still using `onclick=onFormLaunching`. I had purged other uses of `onclick` from the HTML, but apparently not dynamically generated ones from the JS. I looked and didn't find any other problems in the same category.


martinhpedersen (approved):
Nice catch 👏👍

PR #342, opened 2022-04-05
If the developer has Node Version Manager, use it

xylo04:
For some reason, we're stuck on a slightly older (but still supported) version of node 14/npm 6. If we try to run npm commands with the latest LTS node 16/npm 8, `npm install` fails. There's already an `.nvmrc` file in the web directory to signal which version of node is needed, but this causes the `make.bash` file to make sure that version is actually used. Does nothing if the developer does not have nvm installed, they'll just get the warning about "unsupported engine."

### DIFF
--- a/make.bash
+++ b/make.bash
@@ -22,6 +22,10 @@ function install_libax25 {
 }
 function build_web {
 	cd web
+	if [ -d $NVM_DIR ]; then
+	  source $NVM_DIR/nvm.sh
+	  nvm use
+	fi
 	npm install
 	npm run production
 }


martinhpedersen:
😊 It's been a while, but I've implemented similar functionality in other projects using webpack+Go. I will look into it and submit a PR 👍

martinhpedersen:
@xylo04 Maybe we should use [yarn](https://yarnpkg.com/) instead? It just works 🎉

xylo04:
I haven't used yarn before but I'm open to trying that

martinhpedersen:
As I understand it yarn is fully compatible with npm's `package.json`, but differ in that they use different lock files `yarn.lock` vs `package-lock.json`.

> Facebook developed Yarn in 2016 as a replacement for NPM. It was designed to offer more advanced features that NPM lacked at the time (such as version locking) and create a more secure, stable, and efficient product.
(https://phoenixnap.com/kb/yarn-vs-npm)

I've used it in a couple of projects before, but don't have enough experience with either to really give a good comparison of the two. But my experience is that yarn is more stable.

It would be great if you could check it out when you have the time. Should be as easy as `brew install yarn` and then start using it (`yarn install && yarn run production`) 🙂 Let me know how it feels!

martinhpedersen:
Just wanted to add that I don't have strong feelings about this decision. I just noticed that it "just works" with the latest version of yarn, whereas with node we have to pin to an older version 🤷. I think you should decide on this, as you're more experienced in this field 👍

PR #349, opened 2022-04-21
Use ARDOP bandwidth from config as default in web gui

martinhpedersen:

PR #351, opened 2022-05-08
WIP: Add support for Vara HF

MarSik:
This adds support for Vara HF. I am able to complete a Winlink session with standalone Vara running in wine and PAT talking to it directly.

This depends on https://github.com/n8jja/Pat-Vara/pull/17

Also Radio only mode is not yet propagated properly.

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/la5nta/wl2k-go/transport/ardop"
+	"github.com/n8jja/Pat-Vara/vara"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


xylo04:
Thanks for putting this together, but is there a reason to duplicate #280?

MarSik:
Oh, not really, I just did not notice the draft PR in the first place. I was looking for it, but must have overlooked it.

MarSik:
Closed in favor of #280 that is much more complete.

xylo04:
We do need to move #280 forward so thanks for the poke. Martin has some work to do on configurable bandwidths... Oh, I didn't notice he asked for my review on #349, I'd better do that!

PR #353, opened 2022-06-01
Improve dial cancellation

martinhpedersen:
Switch to DialURLContext, which adds generic support for aborting an ongoing dial operation through context cancellation.

It adds dial cancellation for ax25:// and telnet://. The same mechanism is also used to cancel ARDOP dialing (implemented as graceful disconnect), but keeps the old code to provide "dirty disconnect" which is not supported through the context approach and/or the generic conn.Close().

Depends on la5nta/wl2k-go#78.


connect.go
```diff
@@ -131,20 +139,28 @@ func Connect(connectStr string) (success bool) {
 		waitBusy(adTNC)
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
+	dialCancelFunc = func() { dialing = nil; cancel() }
```
xylo04, line 143: Are you sure you want this on clean disconnect, too?

xylo04 (approved):

connect.go
martinhpedersen: Are you referring to the defer dialCancelFunc()? It's a noop to cancel an expired context, in fact it is [required](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/context/context.go;l=230) in order to avoid leaks 🙂

PR #355, opened 2022-06-17
forms: fix to and cc fields

cminyard:
I put this on top of the work by Emil Sit, which still had a couple of issues.  The To and Cc fields still wouldn't get filled in because of a misnamed field in the tokenFieldConfig.  And the split of the fields need to use spaces and commas, too.

### DIFF
--- a/internal/forms/forms.go
+++ b/internal/forms/forms.go
@@ -93,6 +93,8 @@ type FormFolder struct {
 type FormData struct {
 	TargetForm Form              `json:"target_form"`
 	Fields     map[string]string `json:"fields"`
+	MsgTo      string            `json:"msg_to"`
+	MsgCc      string            `json:"msg_cc"`
 	MsgSubject string            `json:"msg_subject"`
 	MsgBody    string            `json:"msg_body"`
 	MsgXML     string            `json:"msg_xml"`
@@ -102,6 +104,8 @@ type FormData struct {
 
 // MessageForm represents a concrete form-based message
 type MessageForm struct {
+	To             string
+	Cc             string
 	Subject        string
 	Body           string
 	AttachmentXML  string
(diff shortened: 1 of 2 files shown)


internal/forms/forms.go
```diff
@@ -1010,6 +1018,8 @@ func (b formMessageBuilder) initFormValues() {
 }
 
 func (b formMessageBuilder) scanTmplBuildMessage(tmplPath string) (MessageForm, error) {
+	var inBody = false
```
martinhpedersen, line 1021: I think this should be moved to just above the loop, to make it easier to understand what it relates to. Also, declaring a bool like this is very unusual in Go. Consider writing it at `inBody := false` or `var inBody bool`.

web/src/js/app.js
```diff
@@ -250,7 +256,7 @@ function initComposeModal() {
     $('#composer').modal('toggle');
   });
   const tokenfieldConfig = {
-    delimiter: [',', ';', ' '], // Must be in sync with SplitFunc (utils.go)
+    delimiters: [',', ';', ' '], // Must be in sync with SplitFunc (utils.go)
```
martinhpedersen, line 259: Is the change from `delimiters` to `delimiter` accidental?

It [looks like](https://sliptree.github.io/bootstrap-tokenfield/) it should be `delimiter`. Please test and verify.

martinhpedersen (commented):
Please read the inline PR comments.

web/src/js/app.js
cminyard: I completely missed this and the above comment, I think some overzealous spam filter drops things from github somtimes.

The change is on purpose.  If you say "delimiter" it will only use one of them.  You have to use "delimiters" to get all of them to work.

I'll fix the non-idiomatic bool and resubmit.

PR #370, opened 2022-08-24
Use VARA HF/FM defaults if undefined in config

martinhpedersen:
This ensures that existing configs without varahf/varafm config gets the default values.


martinhpedersen:
Just to clarify: This won't rewrite the config file, but provide the default values if the config does not define vara config.

PR #372, opened 2022-09-17
Release v0.13.1

martinhpedersen:
Small release, but I think it's best to get the VARA panic fix released as soon as possible.

Changelog:

* Fix panic when using unregistered VARA instances
* Use VARA HF/FM defaults if undefined in config
* Fix case sensitive matching when resolving aux addresses' passwords


xylo04 (dismissed):
Agreed, thank you!

PR #387, opened 2023-01-02
Wait for clear channel (not busy) when using VARA

bseidenberg:
Currently, Pat only implements the busy check for ARDOP. Do the same for VARA. The actual busy/not busy logic is already implemented in Pat-Vara/vara.go in handleCmd().


bseidenberg:
Note: I currently only have remote access to radios to test this, which makes generating the right RF rather difficult. But, it seems straight forward enough given the code.

connect.go
```diff
@@ -153,6 +153,8 @@ func Connect(connectStr string) (success bool) {
 	// Wait for a clear channel
 	switch url.Scheme {
 	case MethodArdop:
+	case MethodVaraFM:
+	case MethodVaraHF:
```
martinhpedersen, line 157: This is not how multiple cases are written in Go, and the code will not work as you expect. See https://go.dev/tour/flowcontrol/10.

The idiomatic way to specify multiple values for the same case would be `case MethodArdop, MethodVaraFM, MethodVARAHF:`.

I haven't checked, but if the VARA transport package implements the `transport.BusyChannelChecker` interface then this could work if you pass the correct TNC reference to the function as well.

martinhpedersen (changes requested):

connect.go
bseidenberg: Ack.

bseidenberg:
Updated

connect.go
```diff
@@ -152,7 +152,7 @@ func Connect(connectStr string) (success bool) {
 
 	// Wait for a clear channel
 	switch url.Scheme {
-	case MethodArdop:
+	case MethodArdop, MethodVaraFM, MethodVaraHF:
 		waitBusy(adTNC)
```
martinhpedersen, line 156: This must also be changed. Passing adTNC here is only correct for ARDOP. We will need to pass a reference to the appropriate TNC connection, which depend on the URL scheme.

martinhpedersen (changes requested):

martinhpedersen (commented):
Looks promising 👍 

The only thing remaining is to test this over some time, and under different conditions. Since we don't provide a method for overriding the "wait busy" state, we depend heavily on a good busy detector.

Do you have any chance of testing this soon, or organize some testing one way or another?

Thanks!

bseidenberg:
Thanks Martin!

I'm away from my radios, but hopefully will be back to them in a few 
more weeks.


Benjamin

martinhpedersen:
@bseidenberg - Did you get a change to test these changes as we discussed? 🙂

bseidenberg:
Thanks for the reminder! I decided to look at this today, and it 
segfaults on a nil pointer exception in transport.go... but I can't 
figure out how it's different from the ARDOP version. I need to brush up 
on my go pointer passing as I debug...stay tuned.

bseidenberg:
OK, I think I understand what's going on...I think. All the other initModem/initTNC() functions are argument-less and directly manipulate the various variables in connect.go for their modems. However initVaraModem() takes a pointer to the modem. That pointer is passed by value as vModem. The vModem local is updated inside initVaraModem() to point to the new modem, but because it's passed by value, the outer "global" of varaHFModem/varaFMModem isn't updated. So, those are always nil, and my code blows up when it tries to dereference them later.

This seems like a bug with the call structure of connect.go, and it being inconsistent... thoughts?

martinhpedersen:
I had a quick look, and I think I've spotted the bug 🙂 

The problem is not in your code. The problem is [`initVaraModem(vModem *vara.Modem, scheme string, conf cfg.VaraConfig)`](https://github.com/la5nta/pat/blob/322113331ae6751522712d4b5f61859ef8ff6176/connect.go#L311). It takes a pointer to vara.Modem, expecting to be able to reassign it. It needs to take a pointer to pointer in order to reassign it like that.

So the global vars `varaHFModem` and `varaFMModem` is never really assigned and is therefore nil when passed to busyWait(). That's probably a bug from the original author that has been this way since VARA support was first introduced.

I can create a branch to fix this, but I don't have the setup to test. Maybe you could help me with that? What we need to confirm is that the fix does not break teardown on subsequent VARA sessions. E.g. make one connection, wait a bit (without stopping Pat) and then do another connection.

bseidenberg:
Yep, happy to test, just shoot me the branch. By the way, if you're on any realtime chat service (IRC?), that'd be great for real-time discussion. Feel free to ping me out of band.

martinhpedersen:
@bseidenberg - See https://github.com/n8jja/Pat-Vara/pull/44.

I have rewritten the way we handle busy channel detection. It all started with #448, and with the refactoring required there I realized we had a good opportunity to add GUI/TUI prompting when the channel is busy.

If/when the pending PR in Pat-Vara is merged, it's a simple one-liner to add VARA busy channel detection 😊 I already have those changes ready, so will probably close the PR once everything is merged.

martinhpedersen:
Closing this in favor of 07391c2b55f51acfd1d3f5870e19def070813250 which adds busy channel detection/handling for both VARA HF and VARAFM, similar to the new approach for ARDOP.

PR #389, opened 2023-01-24
Add support for AGWPE (e.g. Direwolf) over TCP

### LINKED ISSUE #367: Add support for AGWPE over TCP
The AGWPE over TCP protocol is supported natively by Direwolf and AGWPE, allowing the host application to use the soft TNCs built-in AX.25 implementation.

With this, Pat users will be able to do AX.25 talking directly to the open-source software TNC [Direwolf](https://github.com/wb2osz/direwolf), without the hassle of kissattach and axports config. As oppose to today's Linux-only solution, this will be a platform independent feature since Direwolf runs on most operating systems.

See la5nta/wl2k-go#57 for initial discussion on this topic. The transport package is already written.

I'm currently working on integrating it into Pat in a way that (hopefully) will be backwards compatible with existing setups, easily configurable and mostly transparent to the everyday operation in Pat. By introducing the concept of an **AX.25 _engine_**, I hope it will be easy for users to select between multiple AX.25 packet implementations under the same `ax25://` scheme.

martinhpedersen:
This adds support for AGWPE over TCP (closes #367).

It provides a good alternative to the current Linux-only AX.25 transport, by adding a cross-platform solution for interfacing directly with Direwolf and similar software TNCs supporting the AGWPE protocol.

By introducing the concept of an **AX.25 engine**, users will be able to select between multiple AX.25 packet implementations under the same `ax25://` scheme. On Linux builds, the default ax25 engine will be the same as before ("libax25"). On other platforms, AGWPE is the new the default.

To facilitate more advanced use-cases, we're also adding new engine-specific schemes that can be used to reference a specific engine when dialing:
* `ax25+linux://`
* `ax25+agwpe://`
* `ax25+serial-tnc://`

This also deprecates the `serial-tnc://` scheme. Users should instead use `ax25://`, and configure Pat to use the **engine** `serial-tnc`. Alt. use `ax25+serial-tnc://`.

The AGWPE implementation has only been tested with recent versions of Direwolf.


xylo04:
I should have time to look at this tonight!

cfg/config.go
```diff
@@ -251,17 +252,30 @@ type SerialTNCConfig struct {
...
+	Addr string `json:"addr"`
 
-	// Optional beacon when listening for incoming packet-p2p connections.
-	Beacon BeaconConfig `json:"beacon"`
+	// The AGWPE "radio port" (0-3).
+	RadioPort int `json:"radio_port"`
+}
+
+type AX25Config struct {
+	Engine AX25Engine `json:"engine"`
```
xylo04, line 269: It feels slightly wonky to put the engine here, but I suppose it strikes a balance between backward-compatibility and clean config.

usage.go
```diff
@@ -8,20 +8,22 @@ var (
...
-  varahf:     VARA HF TNC
-  varafm:     VARA FM TNC
+  telnet:          TCP/IP
+  ardop:           ARDOP TNC
+  pactor:          SCS PTC modems
+  varahf:          VARA HF TNC
+  varafm:          VARA FM TNC
+  ax25:            AX.25 (Default - uses engine specified in config)
+  ax25+agwpe:      AX.25 (AGWPE/Direwolf)
+  ax25+linux:      AX.25 (Linux kernel)
```
xylo04, line 18: Might be worth calling this by name, `libax25`

xylo04 (approved):
Couple of comments, but no major concerns. 🚀

usage.go
martinhpedersen: Do you mean in the comment or in the scheme? Or both maybe?

Although it is commonly associated with the kernel module, libax25 is actually the name of the C library providing utility functions for interface with the kernel's AX.25 module. It's not really a requirement to utilize the kernel's AX.25 feature and I was actually hoping to port the required functions to Go at some point, meaning we won't use libax25 at all.

But I see your point. Maybe `ax25+linux: AX.25 (Linux kernel aka "libax25")` could work? 🤔

cfg/config.go
martinhpedersen: Yes, I agree. Maybe it's worth revisiting. I'll try a bit more to come up with a better solution. Feel free to share if you have any ideas 🙂 It might be worth designing this without thinking about backward-compatibility, and then try to fix the compatibility with some re-write rules.

cfg/ax25_engine_other.go
```diff
@@ -0,0 +1,6 @@
+//go:build !libax25
+// +build !libax25
```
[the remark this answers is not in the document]
martinhpedersen: Build tags are fun 😀

martinhpedersen:
Just realized that we can add `pactor` as a AX.25 engine as well, since (some of?) those modems support AX.25 🙂 (ping @harenber). Not sure we should prioritize adding that to this PR though, but definitely worth looking into at some point.

cfg/config.go
martinhpedersen: Brainstorming a bit...

# Alternative 1 (current)

I feel we need a common struct to group common AX.25 fields that applies to all engines (beacon and rig config as well as the selected engine). The natural name for this is `ax25`, which we have used for the linux-specific implementation until now.

My main concern with this is that the field name `port` does not reflect the fact that it only applies to the linux engine. 

```json
{
  "ax25": {
    "engine": "linux",
    "port": "wl2k",
    "beacon": {"every": "600", ...},
    "rig": "tmd710"
  },
  "agwpe": {
    "addr": "localhost:8000",
    "radio_port": 0
  }
}
```

# Alternative 2

Following the same pattern as with other modem config, we could move the `port` field into a linux-specific, reserving `ax25` for common fields that applies to all engines.

Today we only have one config field that is specific for `ax25+linux`, but if we in the future get rid of the libax25 dependency (and it's system-wide axport config file) we could see the need to expand with several new fields (`tty`, `serial_baud` etc).

```json
{
  "ax25": {
    "engine": "linux",
    "beacon": {"every": "600", ...},
    "rig": "tmd710"
  },
  "ax25_linux": {
    "port": "wl2k"
  },
  "agwpe": {
    "addr": "localhost:8000",
    "radio_port": 0
  }
}
```

# Alternative 3

We could also just consider renaming the "port" field to better reflect the fact that it only applies to the linux engine. 

```json
{
  "ax25": {
    "engine": "linux",
    "linux_port": "wl2k",
    "beacon": {"every": "600", ...},
    "rig": "tmd710"
  },
  "agwpe": {
    "addr": "localhost:8000",
    "radio_port": 0
  }
}
```

# Alternative 4

We could consider using "packet" instead of "ax25" in both the scheme and config structs, which would create more distance between the scheme and the mode/implementation. This would also be more in line with the wording used in Winlink Express. We could then consider either "linux" or "linux_ax25" as the engine name.

```json
  "packet": {
    "engine": "linux_ax25",
    "beacon": {"every": "600", ...},
    "rig": "tmd710"
  },
  "linux_ax25": {
    "port": "wl2k"
  },
  "agwpe": {
    "addr": "localhost:8000",
    "radio_port": 0
  }
}

```

martinhpedersen: Regarding Alt. 4, I think we should then rename the scheme to `packet://`, `packet+agwpe://`, `packet+linux_ax25://`....

harenber:
Yo.. all PACTOR modems support PACKET RADIO (aka AX.25) - as well as the special 300 Baud ROBUST PACKET mode.

cfg/config.go
martinhpedersen: I'm leaning towards Alt. 2 for the first iteration of this. It's not perfect, but it makes it a lot clearer IMO.

martinhpedersen:
Thanks @harenber! Then I think we should try to add `pactor` as a "ax25 engine" soon after merging this 🎉

dranch:
AX.25 and PACTOR are very different transports and PACTOR is more of a one to one bulk delivery transport where as AX.25 is more of a fully addressable, interactive type protocol.  VARA is more similar to PACTOR in many respects than AX.25

cfg/config.go
martinhpedersen: Pushed a new commit (alternative 2).

martinhpedersen:
> AX.25 and PACTOR are very different transports and PACTOR is more of a one to one bulk delivery transport where as AX.25 is more of a fully addressable, interactive type protocol. VARA is more similar to PACTOR in many respects than AX.25

@dranch - Yes, but many of SCS' modems support AX.25 in addition to PACTOR. That's why it would be nice to allow SCS owners to use their modems for AX.25 as well. This is what we gain by introducing the concept of a selectable "ax25 engine" in Pat 🙂

cfg/config.go
xylo04: Ok, this looks better to me. Thanks!

usage.go
xylo04: I suppose this is ok as-is. I always associated the kernel extension as `libax25` because that's the thing I needed to install to start using it, but it's accurate as you have it.

xylo04 (approved):

cfg/config.go
martinhpedersen: Excellent. I'm pretty happy with it now too 🙂 Thanks for the feedback Chris! 🙇

martinhpedersen:
Rebased. Thanks for the review @xylo04 🙇

PR #390, opened 2023-01-28
Use Github Actions instead of Travis

martinhpedersen:
We keep running out of OSS credits from Travis, and it's a pain having to email them every time to request more.

A osx build step is still TODO. It would be great if we could add that.


martinhpedersen:
I think I've figured out how to do macos and windows tests as well. Will update soon 😊

xylo04 (approved):

PR #391, opened 2023-02-06
Build and run in Docker container

ThreeSixes:
* Create Docker build container
* Update README to include instructions for its use.


martinhpedersen:
Hi there,

Thanks for getting involved in the project 🙂

Please note that we already have a repository dedicated for a dockerized build environment for Pat ([la5nta/pat-build](https://github.com/la5nta/pat-build)).

I'd be happy to review your PR, but first let's discuss the benefits of including this in the repo. The way I see it, it would be better to have a separate repo and maintaining pre-built images @ Docker Hub. We could also include a docker-compose file that will bring up e.g. ARDOP, and/or mount the AX.25 kernel subsystem. What's your take on this?

ThreeSixes:
I should also mention that this isn't purely a build container. It's an HTTP runtime container.

martinhpedersen:
Thanks! I think you have some valid points :)

> I do have a plan to eventually write a KISS-over-TCP transport for Pat

See https://github.com/la5nta/wl2k-go/issues/56 for discussion regarding this.

docker/assets/entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+/app/pat --config /app/config.json \
```
martinhpedersen, line 1: I would prefer that we avoid adding this file.

* We can (to some degree) customize the paths by setting the standard XDG variables `$XDG_DATA_HOME`, `$XDG_CONFIG_HOME` and `$XDG_STATE_HOME`.
* Overriding the callsign with `--mycall` prevents Pat from reading the Winlink password from config, so that should also be avoided IMO. It's better to set mycall in config.

The remaining `pat http` could be set directly in Dockerfile.

README.md
```diff
@@ -57,6 +57,41 @@ This experiment is enabled by default and sessions between two Pat nodes (or oth
...
+```
+docker build . -t latest
+```
+
+### Running the Docker image
+
+Make a directory called `docker/config` and creat a `config.json` file in it. This is the configuration file the Docker container will mount. Make sure the `http_addr` is set to `:8080` in `config.json` otherwise the web interface won't function inside the container.
+
+```
+# Build folder structure.
```
martinhpedersen, line 76: Is this really needed? We automatically create these directories if they don't exist.

README.md
```diff
@@ -57,6 +57,41 @@ This experiment is enabled by default and sessions between two Pat nodes (or oth
...
+mkdir docker/data/logs
+mkdir docker/data/mailbox
+mkdir docker/data/standard_forms
+
+# Run a container and expose the web interface and telnet listener port. Replace `N0CALL` with your actual callsign.
+docker run \
+    --env MYCALL=N0CALL \
+    -p 8080:8080 \
+    -p 8774:8774 \
+    --mount "type=bind,source=$(pwd)/docker/config/config.json,target=/app/config.json" \
```
martinhpedersen, line 87: Wouldn't it be simpler just to mount the single root path?

README.md
```diff
@@ -57,6 +57,41 @@ This experiment is enabled by default and sessions between two Pat nodes (or oth
...
 
+
+## Docker
+
+### Building
+
+To build Pat as a Docker image run the following:
+
+```
+docker build . -t latest
```
martinhpedersen, line 68: `-t pat:latest` maybe?

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+WORKDIR /build
+COPY . /build
+
+# Vet and build
+RUN go vet
+RUN go build
+
+# Runner
+FROM golang:alpine
```
martinhpedersen, line 12: We don't need the full Go compiler to run the app. In fact, I think the much smaller gcr.io/distroless/static will suffice.

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+ENV MYCALL=N0CALL
+
+WORKDIR /app
+
+EXPOSE 8080
+EXPOSE 8774
+
+# Build out directory structure
```
martinhpedersen, line 21: These directories are automatically created by Pat.

Dockerfile
```diff
@@ -0,0 +1,29 @@
+# Builder
+FROM golang:alpine as builder
+
+WORKDIR /build
+COPY . /build
+
+# Vet and build
+RUN go vet
+RUN go build
```
martinhpedersen, line 9: I'd rather we use `make.bash`, as it includes some build info that is very useful when debugging bug reports.

Dockerfile
```diff
@@ -0,0 +1,29 @@
+# Builder
+FROM golang:alpine as builder
+
+WORKDIR /build
+COPY . /build
```
martinhpedersen, line 5: There is no need to specify /build after you have switched with `WORKDIR /build`. `COPY . .` will suffice.

To enable caching of dependencies, I'd like to see this section before `COPY . .`:

```
COPY go.mod go.sum .
go mod download
```

martinhpedersen (changes requested):
Thanks for the PR!

I have some concerns, and some suggestions for simplification. See inline comments 🙂

Dockerfile
martinhpedersen: ... and there is no need to run `go vet` there I think. That is mostly useful to get suggestions on code improvements.

ThreeSixes: > I'd rather we use `make.bash`, as it includes some build info that is very useful when debugging bug reports.

I haven't been able to get this running due to AX.25 dependencies in the Alpine container. It also seems that they don't offer packages for libraries or tools either.

docker/assets/entrypoint.sh
ThreeSixes: > I would prefer that we avoid adding this file.
> 
>     * We can (to some degree) customize the paths by setting the standard XDG variables `$XDG_DATA_HOME`, `$XDG_CONFIG_HOME` and `$XDG_STATE_HOME`.
> 
>     * Overriding the callsign with `--mycall` prevents Pat from reading the Winlink password from config, so that should also be avoided IMO. It's better to set mycall in config.
> 
> 
> The remaining `pat http` could be set directly in Dockerfile.

I'll try to figure out if the `XDG_*` vars work, but it seems like Pat won't run in the container if I don't set `--mycall`. When I set both `--mycall` and configure a callsign in `config.json` it takes the value in `config.json` regardless of the argument set by `--mycall`. I don't know if that's expected.

README.md
ThreeSixes: AFAIK you need to build the local directories on your system before Docker can use them.

Dockerfile
martinhpedersen: The ax.25 dependency should not be needed. Just build with `./make.bash` (without the libax25 sub-command). Should work just fine. If not, please provide the error output so we can debug. Thanks! 🙂

martinhpedersen (changes requested):
See inline comment regarding make.bash. Thanks 🙂

Dockerfile
ThreeSixes: Here's the error when building with make.bash:
```
 => ERROR [builder 7/7] RUN sh make.bash                                                                                                                                                             20.3s
------                                                                                                                                                                                                     
 > [builder 7/7] RUN sh make.bash:                                                                                                                                                                         
#14 0.487 WARNING: No static libax25 library available.                                                                                                                                                    
#14 0.487   Linking against shared library instead. To fix                                                                                                                                                 
#14 0.487   this issue, set CGO_LDFLAGS to the full path of
#14 0.487   libax25.a, or run 'make.bash libax25' to download
#14 0.487   and compile libax25-0.0.12-rc4 in .build/
#14 3.490 Downloading Go dependencies...
#14 3.534 Running tests...
#14 14.25 # github.com/la5nta/wl2k-go/transport/ax25
#14 14.25 /go/pkg/mod/github.com/la5nta/wl2k-go@v0.10.0/transport/ax25/ax25.go:143:16: undefined: DialAX25Context
#14 19.17 ?     github.com/la5nta/pat/cfg       [no test files]
#14 19.17 ?     github.com/la5nta/pat/internal/debug    [no test files]
#14 19.17 ?     github.com/la5nta/pat/internal/directories      [no test files]
#14 19.17 ?     github.com/la5nta/pat/internal/forms    [no test files]
#14 19.17 ?     github.com/la5nta/pat/internal/gpsd     [no test files]
#14 19.17 ?     github.com/la5nta/pat/internal/osutil   [no test files]
#14 19.17 ?     github.com/la5nta/pat/internal/buildinfo        [no test files]
#14 19.17 ?     github.com/la5nta/pat/internal/cmsapi   [no test files]
#14 19.17 FAIL  github.com/la5nta/pat [build failed]
#14 19.20 ok    github.com/la5nta/wl2k-go/catalog       0.009s
#14 19.35 ?     github.com/la5nta/wl2k-go/mailbox       [no test files]
#14 19.35 ?     github.com/la5nta/wl2k-go/rigcontrol/hamlib     [no test files]
#14 19.36 ok    github.com/la5nta/wl2k-go/fbb   0.009s
#14 19.66 ok    github.com/la5nta/wl2k-go/lzhuf 0.311s
#14 19.69 ok    github.com/la5nta/wl2k-go/tests 0.023s
#14 19.70 ok    github.com/la5nta/wl2k-go/transport     0.029s
#14 19.84 ok    github.com/la5nta/wl2k-go/transport/ardop       0.003s
#14 20.04 ?     github.com/la5nta/wl2k-go/transport/telnet      [no test files]
#14 20.04 ok    github.com/la5nta/wl2k-go/transport/ardop2      0.004s
#14 20.04 FAIL  github.com/la5nta/wl2k-go/transport/ax25 [build failed]
#14 20.16 ok    github.com/la5nta/wl2k-go/transport/winmor      0.003s
#14 20.16 FAIL
```

ThreeSixes: I'm thinking we don't have the luxury of a reasonably small compiler because if we must build this way we need AX.25 support in the build container which means Alpine is out.

ThreeSixes: I think based on all the tests I've run so far maybe we should leave this Dockerfile as something that just builds our binary and acts as a runtime container. That has the advantage of not requiring heavy build images on SBCs like Raspberry Pis and of working. I think requiring tests to commit new code makes sense, but I'm wondering if committed and tested code needs to be tested again to build a runtime container. What do you think?

ThreeSixes: As a note: I tested with the golang:latest container as the builder and was able to test when running the libax25 install from make.bash, but after building the binary it doesn't end up working in the runtime container because the binary doesn't run. When I swap back to Alpine as the builder without running the tests everything stays happy and works.

martinhpedersen: Ok, I've done some digging. The problem is that golang:alpine does not include [cgo](https://pkg.go.dev/cmd/cgo) support, and this is not handled very well by neither wl2k-go's ax25 code nor make.bash.

I will push a fix to develop that throws a warning when building under these conditions, but still will produce an acceptable build (without libax25). Should be ok for this setup.

martinhpedersen: Done! Pull in the last commit on develop, and `make.bash` should work just fine with golang:alpine (and distroless) 😄

martinhpedersen: > I think requiring tests to commit new code makes sense, but I'm wondering if committed and tested code needs to be tested again to build a runtime container locally. What do you think?

I don't think it's very much overhead to run these tests. Go code compiles and executes very efficiently 🙂 But we have the option to set `ENV SKIP_TESTS=1` in Dockerfile to have make.bash skip them.

ThreeSixes: I set SKIP_TESTS=1 but the build from script still blows up:
```
=> ERROR [builder 7/7] RUN bash ./make.bash                                                                                                                                        14.8s
------                                                                                                                                                                                    
 > [builder 7/7] RUN bash ./make.bash:                                                                                                                                                    
#14 0.198 WARNING: No static libax25 library available.                                                                                                                                   
#14 0.198   Linking against shared library instead. To fix
#14 0.198   this issue, set CGO_LDFLAGS to the full path of
#14 0.198   libax25.a, or run 'make.bash libax25' to download
#14 0.198   and compile libax25-0.0.12-rc4 in .build/
#14 3.199 Downloading Go dependencies...
#14 3.239 Running tests...
#14 3.239 Skipping.
#14 3.239 
#14 3.239 Building Pat v0.13.1...
#14 12.01 # github.com/la5nta/wl2k-go/transport/ax25
#14 12.01 /go/pkg/mod/github.com/la5nta/wl2k-go@v0.10.0/transport/ax25/ax25.go:143:16: undefined: DialAX25Context
------
executor failed running [/bin/sh -c bash ./make.bash]: exit code: 1
```

ThreeSixes: I have it building. You have to disable the tests and prevent the CGO_* vars from getting set in order for it to build in Alpine.

martinhpedersen: Looks like you missed my [recent changes](https://github.com/la5nta/pat/commit/7bac79726643711379c66566cf5aa3bc485471f7)?

In fact, I see that you've created the PR against `master`. Can you please squash and rebase for `develop` HEAD? This should be mentioned in CONTIRBUTING.md. Sorry about that.

My minimal test Dockerfile works now:

```
➜  pat git:(develop) ✗ cat Dockerfile                                
FROM golang:alpine
WORKDIR /src
RUN apk add --no-cache bash perl git
ADD . . 
RUN ./make.bash

➜  pat git:(develop) ✗ docker build -t pat:latest .
#9 [5/5] RUN ./make.bash
#9 sha256:fffcae5c03e843d632e693c80eccc94620ae4a831ea42ef4e424a1d8c55d220f
#9 0.169 WARNING: CGO unavailable. libax25 (ax25+linux) will not be supported with this build.
#9 0.169 Downloading Go dependencies...
#9 6.640 Running tests...
#9 14.65 ?   	github.com/la5nta/pat/cfg	[no test files]
#9 14.65 ?   	github.com/la5nta/pat/internal/buildinfo	[no test files]
#9 14.65 ?   	github.com/la5nta/pat/internal/cmsapi	[no test files]
#9 14.65 ?   	github.com/la5nta/pat/internal/debug	[no test files]
#9 14.65 ?   	github.com/la5nta/pat/internal/directories	[no test files]
#9 14.65 ?   	github.com/la5nta/pat/internal/forms	[no test files]
#9 14.65 ?   	github.com/la5nta/pat/internal/gpsd	[no test files]
#9 14.66 ?   	github.com/la5nta/pat/internal/osutil	[no test files]
#9 14.67 ok  	github.com/la5nta/pat	0.009s
#9 14.68 ok  	github.com/la5nta/wl2k-go/catalog	0.003s
#9 14.69 ?   	github.com/la5nta/wl2k-go/mailbox	[no test files]
#9 14.69 ?   	github.com/la5nta/wl2k-go/rigcontrol/hamlib	[no test files]
#9 14.71 ok  	github.com/la5nta/wl2k-go/fbb	0.014s
#9 14.94 ok  	github.com/la5nta/wl2k-go/lzhuf	0.245s
#9 14.94 ok  	github.com/la5nta/wl2k-go/tests	0.018s
#9 14.94 ok  	github.com/la5nta/wl2k-go/transport	0.016s
#9 15.01 ok  	github.com/la5nta/wl2k-go/transport/ardop	0.009s
#9 15.05 ok  	github.com/la5nta/wl2k-go/transport/ardop2	0.003s
#9 15.06 ok  	github.com/la5nta/wl2k-go/transport/ax25	0.003s
#9 15.18 ?   	github.com/la5nta/wl2k-go/transport/telnet	[no test files]
#9 15.18 ok  	github.com/la5nta/wl2k-go/transport/ax25/agwpe	0.003s
#9 15.22 ok  	github.com/la5nta/wl2k-go/transport/winmor	0.003s
#9 15.30 
#9 15.30 Building Pat v0.13.1...
#9 16.06 Enjoy!
#9 DONE 16.3s
```

ThreeSixes:
@martinhpedersen this should be ready for review.

Dockerfile
```diff
@@ -0,0 +1,34 @@
+# Builder
+FROM golang:alpine as builder
+
+WORKDIR /build
+
+# Build deps
+RUN apk add --no-cache bash perl git
+
+# Copy source
```
martinhpedersen, line 9: What happened to the `COPY go.mod go.sum .` and `RUN go mod download` section? Did you remove it intentionally?

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+COPY . .
+
+# Build without tests due to AX.25 issues in our build container.
+ENV SKIP_TESTS=1
+ENV NO_AX25=1
```
martinhpedersen, line 14: NO_AX25=1 is no longer needed 🙂

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+WORKDIR /build
+
+# Build deps
+RUN apk add --no-cache bash perl git
+
+# Copy source
+COPY . .
+
+# Build without tests due to AX.25 issues in our build container.
```
martinhpedersen, line 12: I think this comment is misleading. The tests works fine in golang:alpine with the latest modifications in develop`

README.md
```diff
@@ -56,6 +56,31 @@ This experiment is enabled by default and sessions between two Pat nodes (or oth
...
+
+To build Pat as a Docker image run the following:
+
+```
+docker build . -t pat:local
+```
+
+### Running the Docker image
+
+Create a `docker/data/config.json` file in it. This is the configuration file the Docker container will mount. Make sure the `http_addr` is set to `:8080` in `docker/data/config.json` otherwise the web interface won't function inside the container.
```
martinhpedersen, line 72: I think it would be best not to recommend creating `config.json` manually. As long as the directory is valid and mounted, Pat should create this file on first startup with decent defaults (except the required http_addr mod).

runner.sh
martinhpedersen, line 1: I'd prefer to leave this one out. The example `docker run` command in README.md should suffice, shouldn't it?

test.txt
martinhpedersen, line 1: What's this `test.txt`?

martinhpedersen (changes requested):

README.md
ThreeSixes: The web interface doesn't actually work unless you pre-create a `config.json`. The defaults don't work.

martinhpedersen:
@ThreeSixes I looks like you have marked a couple of comments as "resolved" without responding and/or doing the requested changes. Are you still working on this, or are you ready for a new review?

ThreeSixes:
There are some things I have addressed but it's slow going trying to find time to redo a chunk of this to rebase.

martinhpedersen:
Thanks! No rush, take your time 😊

PR #394, opened 2023-02-27
Forms fix for to: and  cc: fields

cminyard:
This is the second try, I fixed the issues you pointed out.

I put this on top of the work by Emil Sit, which still had a couple of issues. The To and Cc fields still wouldn't get filled in because the split of the fields need to use spaces and commas, too.

This is on top of your "develop" branch, not master.

### DIFF
--- a/internal/forms/forms.go
+++ b/internal/forms/forms.go
@@ -94,6 +94,8 @@ type FormFolder struct {
 type FormData struct {
 	TargetForm Form              `json:"target_form"`
 	Fields     map[string]string `json:"fields"`
+	MsgTo      string            `json:"msg_to"`
+	MsgCc      string            `json:"msg_cc"`
 	MsgSubject string            `json:"msg_subject"`
 	MsgBody    string            `json:"msg_body"`
 	MsgXML     string            `json:"msg_xml"`
@@ -103,6 +105,8 @@ type FormData struct {
 
 // MessageForm represents a concrete form-based message
 type MessageForm struct {
+	To             string
+	Cc             string
 	Subject        string
 	Body           string
 	AttachmentXML  string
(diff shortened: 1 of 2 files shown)


martinhpedersen:
Thanks for the new PR @cminyard 😄

internal/forms/forms.go
```diff
@@ -1011,6 +1019,8 @@ func (b formMessageBuilder) initFormValues() {
 }
 
 func (b formMessageBuilder) scanTmplBuildMessage(tmplPath string) (MessageForm, error) {
+	inBody := false
```
martinhpedersen, line 1022: Minor detail, but I would rather this be defined just before the `for` loop since this is where it is used.

internal/forms/forms.go
```diff
@@ -1053,8 +1067,14 @@ func (b formMessageBuilder) scanTmplBuildMessage(tmplPath string) (MessageForm,
...
-		} else {
+		} else if strings.HasPrefix(lineTmpl, "To:") {
+			retVal.To = strings.TrimPrefix(lineTmpl, "To:")
+		} else if strings.HasPrefix(lineTmpl, "Cc:") {
+			retVal.Cc = strings.TrimPrefix(lineTmpl, "Cc:")
+		} else if inBody {
 			retVal.Body += lineTmpl + "\n"
+		} else {
+			log.Printf("skipping unknown template line: '%s'", lineTmpl)
 		}
```
martinhpedersen, line 1078: I think this would be easier to read if was written as switch, like so:

```go
		switch {
		case strings.HasPrefix(lineTmpl, "Subject:"):
			retVal.Subject = strings.TrimPrefix(lineTmpl, "Subject:")
		case strings.HasPrefix(lineTmpl, "To:"):
			retVal.To = strings.TrimPrefix(lineTmpl, "To:")
		case strings.HasPrefix(lineTmpl, "Cc:"):
			retVal.Cc = strings.TrimPrefix(lineTmpl, "Cc:")
		case inBody:
			retVal.Body += lineTmpl + "\n"
		default:
			log.Printf("skipping unknown template line: '%s'", lineTmpl)
		}
```

martinhpedersen (approved):
Had a second read through this, and notices a couple of possible improvements on the code style to better align with the rest of the codebase. Nothing important, and I can easily fix this if you prefer.

I'll wait a couple of days before merging in case you want to fix this 🙂

martinhpedersen:
Thanks for this @cminyard, @sit 😄

cminyard:
On Sun, Apr 16, 2023 at 05:16:38AM -0700, Martin Hebnes Pedersen wrote:
> Thanks for this @cminyard, @sit 😄 

Thanks for merging this, and thanks for a great program.

PR #395, opened 2023-03-28
Get web build working

xylo04:
I can't currently build this at all because `node-sass` is old and requires python2. The replacement `sass` does not.


martinhpedersen:
Thanks 😄 

Do you have any idea why the logo png has been changed by this PR? Do we have a image compressor in our build pipeline with an unstable output maybe? 🤔

xylo04:
> Do you have any idea why the logo png has been changed by this PR? Do we have a image compressor in our build pipeline with an unstable output maybe? 🤔

I think there is an image compressor in the pipeline. It's not usually unstable, but perhaps this particular change caused a change in the compressor as well. I did a little bit of examining, it's basically the same size and content. If it ends up being nondeterministic I'll see what I can do about excluding it from commits.

martinhpedersen (approved):
Thanks 😊

PR #397, opened 2023-04-04
Re-init Form Templates list after updating

### LINKED ISSUE #396: Updating Forms data requires a page reload in http mode
Is there a way to force this as part of the update process?

martinhpedersen:
Resolves #396 without having to invalidate all state by reloading the web app, I hope 🙂 

I'm not a Forms user myself, but I did a couple of quick tests and it looks promising 🤷


xylo04 (approved):
Thanks Martin!

PR #411, opened 2023-05-21
web: Add all the ax25 transports to the transport menu

cminyard:
So they can be individually selected and used from the web page.

### DIFF
--- a/web/src/index.html
+++ b/web/src/index.html
@@ -199,17 +199,16 @@
                           data-width="100%"
                           id="transportSelect"
                         >
-                          <optgroup label="transports">
-                            <option value="ardop">ARDOP</option>
-                            <option value="ax25">AX.25</option>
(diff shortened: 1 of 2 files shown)


martinhpedersen (approved):
Thanks! I forgot about this 🙂 

I'm a bit concerned that novice users may find it confusing when they are prompted with this many different ax25 options.

So i'd like to make a few adjustments:
* Move the `ax25+...` options to a separate option group labeled "advanced".
* Remove the legacy serial-tnc option (same as the new ax25+serial-tnc).

martinhpedersen:
Something like this:

image

cminyard:
On Sun, May 21, 2023 at 02:34:50AM -0700, Martin Hebnes Pedersen wrote:
> Something like this:
> 
> image

Yes, that's better.  Thanks.

martinhpedersen:
Thank you 😊

PR #422, opened 2023-07-31
Only attach Forms XML if a viewer file is defined

### LINKED ISSUE #420: Forms and file attachments
Consider the following form definition file:

```
Form: FFWN.html
Subject: FFWN
Readonly:True
Msg:
<var Callsign>,<var Name>,<var City>,<var Region>,<var Country>,<var Username>,<var Connection>
<var Question>
```

(see https://github.com/Mihara/ffwn for more details)

RMS Express, when working with this file, produces an email *without* an XML file attachment, and puts the form data into the message body as expected. Pat, however, insists on attaching the XML file anyway.

There is no whiff of a formal specification of how forms *should* work -- I didn't find one myself, at least -- but I think this goes against the defacto standard.

martinhpedersen:
Resolves #420

### DIFF
--- a/cli_composer.go
+++ b/cli_composer.go
@@ -332,8 +332,10 @@ func composeFormReport(ctx context.Context, args []string) {
 
 	msg.SetBody(formMsg.Body)
 
-	attachmentFile := fbb.NewFile(formMsg.AttachmentName, []byte(formMsg.AttachmentXML))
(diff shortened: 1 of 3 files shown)


xylo04:
Sorry I missed this. It looks good.

martinhpedersen:
No problem 😊 Thanks 😀

PR #423, opened 2023-10-03
fix tests and compilation with `TAGS=libhamlib`

as3ii:
**Note: this requires the merge of https://github.com/la5nta/wl2k-go/pull/85**

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,20 @@
+pat (0.15.0) stable; urgency=medium
+
+  * Restore previous connect parameters from browser's local storage
+  * Add missing AX.25 schemes to connect modal's transport dropdown
+  * Fix clearing of To/Cc fields after message is posted to outbox
+  * Fix alignment of connect modal input fields
+  * Improve the dirty disconnect feature
+  * Add deprecation warning for newly deprecated config options
+  * Remove support for previously deprecated config options
+  * AGWPE: Add support for QtSoundModem
+  * AGWPE: Wait for modem ack on dial cancellation
+  * VARA: Add support for inbound (P2P) connections
+  * VARA: Improved throughput, various bug fixes and other improvements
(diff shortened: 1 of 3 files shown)


as3ii:
I don't know precisely if this is related to the use of libhamlib, and if this is an issue in pat or in wl2k-go, but sometimes randomly pat starts eating a lot of cpu power. The only way I was able to reproduce consistently this bug was manually disconnecting and reconnecting the usb serial interface to the radio.
This is what `perf` recorded and found when this happens:
perf

Tell me if I can do anything to help debug this problem

riglist.go
```diff
@@ -10,7 +10,7 @@ package main
 import (
 	"fmt"
 	"strings"
-
+	"context"
```
martinhpedersen, line 13: I believe we're missing a white space here. Did you [go fmt](https://go.dev/blog/gofmt) the code?

martinhpedersen:
> I don't know precisely if this is related to the use of libhamlib, and if this is an issue in pat or in wl2k-go, but sometimes randomly pat starts eating a lot of cpu power.

Thanks! I don't think i have seen this when Pat is compiled without `TAGS=libhamlib`. The feature is rarely used and not enabled in the official releases, so I don't think I'll prioritize debugging this any time soon. Feel free to debug further if you feel like it 🙂 

PS: Thanks for the PR! ❤️

riglist.go
as3ii: added a commit after `go fmt`. It just reordered the entries

martinhpedersen:
I've merged the wl2k-go changes and pulled them into Pat's develop branch. Please rebase this PR on top of `develop`. Thank you 🙂

martinhpedersen:
Ooops! I did something wrong while trying to rebase this, and since the PR was automatically closed I no longer have permission to push to your branch.

I'll merge this manually to develop.

Thanks for the PR 😄

martinhpedersen:
Merged (b62dbc787b294e63e30da568ac3ea423f2abbc97).

PR #424, opened 2023-10-07
Build and run in Docker

martinhpedersen:
Replaces draft PR #391 which is stalled. FYI @ThreeSixes.

This adds a Dockerfile for building a minimalistic Docker image with Pat. It also adds a docker-compose file as a starting point for those that wishes to experiment with this kind of setup. This is my first attempt of creating a Docker image for public release, following the best practices to the best of my knowledge.

In the future I'd like to configure a Github Actions pipeline for pushing these images to dockerhub. I'd also like add a new repo containing a Docker Compose setup including ARDOPc, Direwolf. It will probably be useful mostly for Linux users, since Docker Desktop on mac/windows does not support sound devices. But for Linux users, I think this could prove to be an easy and convenient way to deploy a robust and reproducable software suite for Winlink. I've already started experimenting with such a system on an old x86 laptop. It would be great to include VARA HF/FM also, but there are licensing issues to consider.


ThreeSixes:
Thanks for taking this over. I've had some life stuff.

Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN go mod download
 ADD . .
 RUN go build -o /src/pat
 
-FROM gcr.io/distroless/static
+FROM scratch
```
martinhpedersen, line 9: Regarding this change.

> Statically compiled applications (Go) that do not require libc can use the gcr.io/distroless/static image, which contains:
> 
> ca-certificates
> A /etc/passwd entry for a root user
> A /tmp directory
> tzdata

So I'm going to look into if we need any of these.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN go build -o /src/pat
+
+FROM scratch
+LABEL org.opencontainers.image.source=https://github.com/la5nta/pat
+LABEL org.opencontainers.image.description="Pat - A portable Winlink client for amateur radio email"
+LABEL org.opencontainers.image.licenses=MIT
+COPY --from=builder /src/pat /bin/pat
+USER 65534:65534
+WORKDIR /app
+ENV XDG_CONFIG_HOME=/app
```
xylo04, line 16: I double-checked, these are the three XDG directories we're using.

xylo04 (approved):
Looks good as a first pass!

martinhpedersen:
Thank you for the review 😄

martinhpedersen:
I pushed Github Actions workflow for pushing to Docker Hub. Do you mind taking a look at this as well? It seems to be working 🙂 

One of the test runs: https://github.com/la5nta/pat/actions/runs/6523720316/job/17714636579

xylo04:
> I pushed Github Actions workflow for pushing to Docker Hub. Do you mind taking a look at this as well? It seems to be working 🙂
> 
> One of the test runs: https://github.com/la5nta/pat/actions/runs/6523720316/job/17714636579

Looks good! I see it's already published at https://hub.docker.com/r/la5nta/pat.

One curious side note, your callsign/account name seems to be getting replaced in the Actions build logs, e.g. [this](https://github.com/la5nta/pat/actions/runs/6523720316/job/17714636579#step:7:123). Doesn't seems like a problem, just weird.

For my education, I see on Docker Hub there are several digests for different architectures, but it doesn't look like you're having to explicitly build them all in the build step. Do you get containers for all of the different architectures "for free?"

xylo04:
Also, I just noticed you're targeting the `master` branch instead of `develop`, I presume that's on purpose since this packaging-related instead of feature-related.

xylo04 (approved):

martinhpedersen:
> One curious side note, your callsign/account name seems to be getting replaced in the Actions build logs, e.g. [this](https://github.com/la5nta/pat/actions/runs/6523720316/job/17714636579#step:7:123). It doesn't seem like a problem, just weird.

Yes, I thought so too. I have no idea why it's masked 🤷 


> For my education, I see on Docker Hub there are several digests for different architectures, but it doesn't look like you're having to explicitly build them all in the build step. Do you get containers for all of the different architectures "for free?"

Yes! Well, almost. If you look closely in the workflow definition you'll see I've added a list of platforms. But basically that's the magic of Docker's Buildx (https://github.com/docker/setup-buildx-action) 😄 It's very neat!


> Also, I just noticed you're targeting the master branch instead of develop, I presume that's on purpose since this packaging-related instead of feature-related.

Yes, you assumed correct 🙂 I figured it's ok to commit this directly to master and not tagging a new version, since it does not affect the app itself.

PR #425, opened 2023-10-15
Forms date formatting

martinhpedersen:
I discovered an error in the time formatting strings we use in the forms package, so I wanted to write some tests and fix this.

As it turns out, the error did not affect the outputted strings. It only manifests when outputting datetime strings not in UTC, which none of these does. Anyway, I've added some tests and fixed the layout strings 🙂


xylo04 (approved):
Interesting bug, and nice fix! :)

PR #426, opened 2023-10-16
Update node version and transitive deps

xylo04:
This updates Node and NPM to the latest Long Term Supported versions. Node is imminently upgrading v20 to LTS, so we should at least get this far.


martinhpedersen:
Thanks :D

While I was testing this I saw a very strange bug. One of the messages I received was truncated before I got to reading it. Either that, or the file was written as empty (0 bytes) without causing any errors. The output view (web) refreshed after CMS exchange, and then it errored out on io.EOF when trying to read the message.

I was testing on my Android phone, so I have a feeling this is not related to the changes in the PR. Nevertheless, having a message truncated after being ACKed (and thus deleted from CMS) is a bug worth investigating. The message is gone. I've read through the Go and JS code multiple times now, and can't understand how this could have happened.

The only thing I can think of related to this PR is the `setRead` JS function which triggers a re-write of the message on disk. Might be a race related to that, but I can't understand why/how it was triggered if that's the case.

martinhpedersen:
So I did some more digging, and found a case where we would end up writing an empty .b2f file that could explain the bug I observed.

If the B2F message we receive is a headerless message (or a message starting with the header delimiter), we end up not reporting any error and just writing an empty message to disk. The empty file written to disk as a result will produce the error I observed when testing this PR. Unfortunately, since the buggy message was lost I can't confirm my theory.

I think it's safe to say the bug isn't related to this PR. I'll merge this asap, but first some sleep 🛌 🙂

martinhpedersen (approved):

PR #429, opened 2023-11-04
Support config overrides using env variables

martinhpedersen:
Override configuration values by setting PAT_* environment variables. This is handy when deploying a containerized/scripted setup.

Syntax examples:
* PAT_MYCALL=LA5NTA
* PAT_ARDOP_ADDR="host:port"

### DIFF
--- a/config.go
+++ b/config.go
@@ -11,15 +11,26 @@ import (
 	"path"
 	"strings"
 
+	"github.com/kelseyhightower/envconfig"
 	"github.com/la5nta/pat/cfg"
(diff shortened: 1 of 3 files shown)


xylo04 (approved):
We should couple this PR with wiki information about what these env vars look like.

martinhpedersen:
Thanks for the review 🙂  We don't have a dedicated wiki page regarding configuration of Pat, but I added a (very) small section here: https://github.com/la5nta/pat/wiki/The-command-line-interface#environment-variables. In general I think configuration deserves it's own wiki page.. its a big topic 😅

Issue #114, opened 2017-12-18
Packet Script Support


ATCUSA:
Our local area winlink connection uses a vhf node that connects to the uhf winlink gateway. Using RMS express you can handle this by using a connect script instead of the digipeater mode. The script would look like the following( anything after // is a comment and not used in the actual rms script.):
```
C K7LL-7 // Connects to Node
Connected // Looks for this keyword to be returned from the node. Not case sensitive.
C K7LL-10 // Tells the node to connect to the Winlink gateway
Connected
```
We are trying to move away from rms express over to pat for our local emergency communications team.

martinhpedersen:
Hi - thanks for the proposal.

We got a similar feature request on [pat-users a while back](https://groups.google.com/forum/#!topic/pat-users/S3wfxTXn93c). Maybe you could discuss this further with Mike to ensure that both use-cases can be handled by this feature request?

The syntax from your example script seems to be somewhat ambiguous IMHO, with even/odd lines indicating send/expect. Is this a common syntax for node traversal? Maybe we could come up with something a tittle more flexible and robust(?).

From the top of my head:
1. An embedded scripting language [Lua?](https://github.com/yuin/gopher-lua)
2. Exec a process and connect stdin/stdout to the AX.25 stream (Use the returned status code to signal success/failure)
3. [Go's new plugin support](https://jeremywho.com/go-1.8---plugins/)
4. Define our own syntax. E.g. *(oh, please stop me...)*
    ```
        send "C K7LL-7\r\n"  // Connect to node
        wait "*[Cc]onnect.*" // Wait for line that matches this regular expression
    ```
5. Keep it simple (your proposal)

I think no 2. (exec) probably would be the most realistic solution with regards to implementation complexity and usage flexibility. With that approach, you could even plug in a custom parser to support what ever syntax you prefer.

What do you think?

ATCUSA:
I think you are on to a good idea. I am pretty sure the idea is similar to what the other user was trying to describe. Basically instead of connecting through a digipeater we connect to a node and then have to tell that node to connect to the gateway. I can't say for certain that it is overly common, however, this is how RMS Express handles it. I think that number 2 is a good approach, but I also like the idea of going with the new go plugins. Please let me know if you need me to further clarify or if you need any help with testing things out since. Thanks again for all the work and time you put into this project.

Austin




[...]

PR #432, opened 2023-11-23
Add connection prehook for scripted node traversal

martinhpedersen:
Adds low-level connection prehook for all transports, implemented by spawning a user-defined process communicating with the remote station over stdio.

This enables scripted node traversal (#114) at the lowest possible level. To make it more practical and user-friendly, we need to add a layer on top. Not everyone is comfortable writing bash scripts/programs able to do this, so we need a simplified language with just enough flexibility to express the command/response sequences commonly used in packet node traversal.

------

TODO
- [ ] Write documentation
    - [x] New connect URI query parameters (wiki + `pat connect help`)
    - [ ] Describe the protocol between Pat and the spawned process (wiki)

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -184,6 +184,13 @@ func Connect(connectStr string) (success bool) {
 	dialing = url
 	websocketHub.UpdateStatus()
 
+	if exec := url.Params.Get("prehook"); exec != "" {
+		if err := VerifyPrehook(exec); err != nil {
+			log.Printf("prehook invalid: %s", err)
+			return
+		}
+	}
+
 	log.Printf("Connecting to %s (%s)...", url.Target, url.Scheme)
 	conn, err := transport.DialURLContext(ctx, url)
 
--- a/connect.go
+++ b/connect.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/la5nta/pat/cfg"
 	"github.com/la5nta/pat/internal/debug"
+	"github.com/la5nta/pat/internal/prehook"
 
 	"github.com/harenber/ptc-go/v2/pactor"
 	"github.com/la5nta/wl2k-go/transport"
@@ -185,7 +186,7 @@ func Connect(connectStr string) (success bool) {
 	websocketHub.UpdateStatus()
 
 	if exec := url.Params.Get("prehook"); exec != "" {
-		if err := VerifyPrehook(exec); err != nil {
+		if err := prehook.Verify(exec); err != nil {
 			log.Printf("prehook invalid: %s", err)
 			return
 		}
@@ -211,7 +212,7 @@ func Connect(connectStr string) (success bool) {
 
 	if exec := url.Params.Get("prehook"); exec != "" {
 		log.Println("Running prehook...")
-		prehookConn := NewPrehookConn(conn, exec, url.Params["prehook-param"]...)
+		prehookConn := prehook.Wrap(conn, envAll(), exec, url.Params["prehook-param"]...)
 		if err := prehookConn.Wait(ctx); err != nil {
 			conn.Close()
 			log.Printf("Prehook script failed: %s", err)
--- a/connect.go
+++ b/connect.go
@@ -9,11 +9,13 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/la5nta/pat/cfg"
+	"github.com/la5nta/pat/internal/buildinfo"
 	"github.com/la5nta/pat/internal/debug"
 	"github.com/la5nta/pat/internal/prehook"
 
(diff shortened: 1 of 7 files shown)


martinhpedersen:
... We should probably define a location for these scripts and append that to the spawned process' `PATH`.

internal/prehook/prehook.go
```diff
@@ -0,0 +1,175 @@
...
+type Conn struct {
+	net.Conn
+	br *bufio.Reader
+}
+
+// Verify returns nil if the given script file is found and valid.
+func Verify(file string) error { _, err := lookPath(file); return err }
+
+func lookPath(file string) (string, error) {
+	// Look in our custom location first
```
xylo04, line 50: This works, but in an earlier comment you described trying to append `prehooks` to the process' `PATH`. If that worked the way I assumed, our `prehooks` dir would just be the first place `LookPath` would explore, and you wouldn't have to do two checks. I'm guessing that didn't work out?

connect.go
```diff
@@ -202,6 +212,26 @@ func Connect(connectStr string) (success bool) {
 		return
 	}
 
+	if exec := url.Params.Get("prehook"); exec != "" {
+		log.Println("Running prehook...")
+		script := prehook.Script{
+			File: exec,
+			Args: url.Params["prehook-arg"],
```
xylo04, line 219: I suppose I could just try this, but I can also ask you about your intent: `url.Params["prehook-arg"]` gives back an array of args. If I have multiple args I need to pass to the script, do I make multiple URL params with the same key?

E.g. if I want to invoke the prehook script with:

```
my-prehook.sh --bool-arg --arg2=value
```

then I encode that in the URL as

```
ax25:///K7LL-5?prehook=my-prehook.sh&prehook-arg=--bool-arg&prehook-arg=--arg2=value
```

usage.go
```diff
@@ -33,6 +33,12 @@ path:
 params:
   ?freq=        Sets QSY frequency (ardop and ax25 only)
   ?host=        Overrides the host part of the path. Useful for serial-tnc to specify e.g. /dev/ttyS0.
+  ?prehook=     Sets an executable middleware to run before the connection is handed over to the B2F protocol.
```
xylo04, line 36: High-level question: since we are making a connection pre-hook for initializing connections, does that now require that there's a post-hook for tearing the connection down cleanly? Maybe the proposed use case, AX.25 packet, is smart enough to tear down traversal links.

internal/prehook/prehook.go
```diff
@@ -0,0 +1,175 @@
...
+func (p *Conn) Read(b []byte) (int, error) { return p.br.Read(b) }
+
+// Execute executes the prehook script, returning nil if the process
+// terminated successfully (exit code 0).
+func (p *Conn) Execute(ctx context.Context, script Script) error {
+	execPath, err := lookPath(script.File)
+	if err != nil {
+		return err
+	}
+	cmd := exec.CommandContext(ctx, execPath, script.Args...)
```
xylo04, line 82: How does the pre-hook script communicate its progress? Is exit code 0 the main/only way? I could see eventually expanding it to where the connection script can signal that the connection is ready, then Pat can signal when B2F is done and let the script tear down the connection. But I don't think that's necessary for v1.

xylo04 (approved):

internal/prehook/prehook.go
martinhpedersen: Yes, that would be much easier. I thought the issue was that Go's standard library didn't provide a portable way of modifying $PATH. The syntax differs between Windows and linux/darwin.

.. However, while writing this comment I discovered [os.PathListSeparator](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/os/path_unix.go;l=11) 😄 I'll give it another try! Thanks 🙂

connect.go
martinhpedersen: Yes, that's correct 🙂 As an alternative we could consider a single argument with a separator. However, then we would also need to implement escaping in case the separator should be passed to the script.

usage.go
martinhpedersen: That's a very good question! 🤔 

I would imagine that all connections would be teared down once one of the links are disconnected. Otherwise, you'd run the risk of dangling connections in case the dialer looses it's connection. But this is definitely something to investigate.

internal/prehook/prehook.go
martinhpedersen: Yes, exit code 0 == success. Any other exit code is assumed to be a failure, and the connection is closed with an error presented to the user.

I see your point. We could maybe use a signal like `SIGCONT` for this, to keep stdio as transparent as possible. We would also need another mechanism to communicate that the connection is ready for B2F, as this is currently implemented by the child process exiting. But yes, maybe not necessary for v1.

usage.go
martinhpedersen: We could check to see if Winlink Express provides such a feature. If they don't, I would imagine it's not very common 😁

PR #435, opened 2023-12-12
Integrated grid square calculations with GPSd

n8jja:
1. Moved the position checking inside a Goroutine that runs every 5 minutes.
2. Created `updateGridIfNeeded` function, it checks the grid location in `config.Locator` and updates it if necessary.
3. Expanded the functionality of the GPSd TPV object to include the grid square as one of its values.
4. Consolidated all functionality for Maidenhead calculations inside GPSd for simplicity.


internal/gpsd/objects.go
```diff
@@ -27,6 +28,23 @@ type TPV struct {
...
+	}
+
+	lon, err := t.Lon.Float64()
+	if err != nil {
+		return err
+	}
+
+	point := maidenhead.NewPoint(lat, lon)
+	t.GridSquare, err = point.GridSquare()
+	return err
```
martinhpedersen, line 47: This is no longer needed I believe? `conn.GetGridSquare()` is not using the TPV's GridSquare, but calculates from LatLon on the fly.

main.go
```diff
@@ -387,6 +390,48 @@ func main() {
...
+	// get grid square from position
+	currGridSquare := pos.GridSquare
+
+	if currGridSquare != config.Locator {
+		// update config
+		config.Locator = currGridSquare
+		// write config
+		if err := WriteConfig(config, fOptions.ConfigPath); err != nil {
+			log.Printf("Unable to write config: %s", err)
+		}
```
martinhpedersen, line 431: I don't think rewriting the actual config file is the best idea. It is not a lossless operation, as the user might have defined keys and values that are not part of the JSON spec (e.g. "comments"). Custom formatting will also be lost, and possibly also ordering of aliases etc. It also has the potential of corrupting the file if we for some reason the operation fails in the middle of writing (full disk, power loss++).

I guess it's useful to retain the "last known" position when restarting the app. Otherwise, the user would have to wait the 5 min interval before the locator field is updated again

As I see it, we have two options:
1. Persist the "last known position" in a separate JSON file, and use that as a initial value (if present).
2. Block on startup until we get a fresh position from GPSd.

I think the latter might be a better solution, as it guarantees running with `--locator-from-gpsd` uses a value from GPSd at all times.

main.go
```diff
@@ -387,6 +390,48 @@ func main() {
 
 	// Start command execution
 	cmd.HandleFunc(ctx, args)
+
+	// Go routine for checking the GPS
+	go func() {
+		if fOptions.UpdateGrid {
+			for range time.Tick(time.Minute * 5) {
+				updateGridIfNeeded()
+			}
```
martinhpedersen, line 399: What's the reason for polling once per 5 minute? Wouldn't it be even better to continuously keep the locator field up-to-date by subscribing (Watch(true)) and call Next() in an endless loop? 🤔

main.go
```diff
@@ -387,6 +390,48 @@ func main() {
...
+	conn, err := gpsd.Dial(config.GPSd.Addr)
+	if err != nil {
+		log.Printf("GPSd daemon: %s", err)
+		return
+	}
+	defer conn.Close()
+	conn.Watch(true)
+
+	// get next position from gpsd
+	pos, err := conn.NextPos()
```
martinhpedersen, line 416: I think using `conn.GetGridSquare()` instead would simplify this PR a lot 🙂

internal/gpsd/gpsd.go
```diff
@@ -235,3 +243,24 @@ func errUnexpected(err error) error {
 	}
 	return err
 }
+
+// GetGridSquare provides function for getting updated position and calculating the grid square.
+func (c *Conn) GetGridSquare() (string, error) {
```
martinhpedersen, line 248: I believe `NextGridSquare()` would be a better name for this, to better harmonize with `Next()` and `NextPos()` 🙂

martinhpedersen (changes requested):
Thanks! I think we're almost there 😄 

It appears to me like this PR contains traces of multiple approaches we discussed in the email.

In my mind, this is what we discussed:

Simplest (requires the least changes):

```go
func (p Position) GridSquare() (string, error) {
  return maidenhead.NewPoint(p.Lat, p.Lon).GridSquare()
}
```

...  alternatively add field Position.GridSquare and enrich in `func (t TPV) Position() Position` ignoring the error.
```go
func (t TPV) Position() Position {
  ...
  // The error is ignored as it's always nil for valid coordinates.
  p.GridSquare, _ := maidenhead.NewPoint(p.Lat, p.Lon).GridSquare()
  return p
}
```

And finally:

```go
func (c *Conn) NextGridSquare() (string, error) {
  pos, err := c.NextPos()
  if err != nil {
    return "", err
  }
  return maidenhead.NewPoint(lat, lon).GridSquare()
}
```

The latter is already implemented in this PR, but not used.

Which of the options do you think is best?

In addition I have some concerns regarding the way you persist the updated `locator` to disk.  Please see in-line comments.

internal/gpsd/gpsd.go
```diff
@@ -50,6 +52,12 @@ type Conn struct {
 	closed       bool
 }
 
+type GPSdClient interface {
+	Dial(addr string) (GPSdClient, error)
+	Watch(enable bool)
+	Next() (*TPV, error)
+}
+
```
martinhpedersen, line 60: This is unused? If so, please omit it 😊

kd5pqj: I am looking at storing the Locator in the user.json for the multi-user reloader.  Like KM4ACK's user swap by config swap
in the PAT Menu 12.  GPSd is great and I have one pi feeding 4 pi's locator.  only have a 20 minute gap every other month.
in mailbox/CallSign/user.json is (CALLSIGN, Name, TAC), Locator (EL29ER27),  WinLink/user (Password), Roll (Owner, Radio Op, Non-Ham), On Duty(00-23), NTS (routing itu-r2, NOAM, USA, TX, STX,)  HW ACCESS (Telnet, Wifi2, Wifi5, Ardop, ### Ax25, VaraHF, VaraFM, VaraSat)
Just have to get the Variables loaded at the right time for each Network http connection session or postoffice push/pull.
This lets windows and mac users to get multi-user.... Roll->Multi-user->PostOffice->NTS
Herbert KD5PQJ

internal/gpsd/objects.go
n8jja: Hi Martin!

You are correct...it is an artifact that I forgot to remove.  Thanks!

main.go
```diff
@@ -214,6 +215,7 @@ func optionsSet() *pflag.FlagSet {
 	set.BoolVarP(&fOptions.SendOnly, "send-only", "s", false, "Download inbound messages later, send only.")
 	set.BoolVarP(&fOptions.RadioOnly, "radio-only", "", false, "Radio Only mode (Winlink Hybrid RMS only).")
 	set.BoolVar(&fOptions.IgnoreBusy, "ignore-busy", false, "Don't wait for clear channel before connecting to a node.")
+	set.BoolVar(&fOptions.UpdateGrid, "gridsquare-update", true, "Automatically update the maidenhead grid square from the gpsd daemon.")
```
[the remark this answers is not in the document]
n8jja: Hi Martin!

Yes, I will change to align with the existing terminology. It makes the most sense to keep things consistent.

main.go
n8jja: Thanks Martin...I agree the latter would be the better option.

main.go
n8jja: My thought was that while most people will be mostly static...i.e. once they get the first fix from the GPS, they won't need it anymore.  However, for true mobile operation, it might be required to get a new grid square at a regular interval.  

I picked 5 minutes at random for testing, but I think long term, this might be something we allow the user to enable/disable and set the interval as needed. If you were in a moving car for example, 5 minutes might be a good interval, but on a sailboat, it might need be only once every few hours.  

I'll disable the loop for now.

internal/gpsd/gpsd.go
n8jja: Absolutely!

main.go
n8jja: After some refreshing on this...won't the gaps.Dial() and conn.NextPos() already blocking?

PR #436, opened 2023-12-23
Hardening related to resolving Forms files

martinhpedersen:
Improved error handling, more efficient search and make sure we don't escape the Forms root directory. Also adds debug logging and fixes some reply templates we missed before.


xylo04 (approved):
Nice improvements!

martinhpedersen:
Thanks for the review Chris 😄 

I did some more testing over the holiday and spotted some issues related to reading template files outside the Standard Forms directory (i.e. with `pat composeform --template ~/foo/bar/.txt`). I also identified a couple more weak points and did some more refactoring.

I'll commit further changes to the PR over the next days. Will let you know when I believe it's ready for a new review 🙂

martinhpedersen:
Ready for a new review if you don't mind 🙂 

I ended up doing a lot more refactoring, and at the same time squashed a couple of bugs. There is still plenty of work to be done, especially regarding support for txt-only templates. However, I think think it's best to do this in smaller increments som it's easier to do reviews and keep track of the changes.

PS: I starting working on this because NRRL (Norwegian Radio Relay League) had a x-mas exercise using templates this week, and I got a lot of questions from Norwegian hams struggling to get things working properly. I realize I need to invest more time in the Forms feature to get it up to the current "standards".

PR #438, opened 2024-01-01
Add support for Message Templates (templates without HTML Forms)

### LINKED ISSUE #375: Support Message Templates (without HTML Forms)
Ref https://github.com/la5nta/pat/issues/374#issuecomment-1262670594.

There is a difference between Message Templates and HTML Forms. Our `package forms` ignores templates without HTML Forms, and does not support the full scripting language commonly used with pure Message Templates.

martinhpedersen:
This adds supports for templating features typically used in text-only templates:
* `<Ask ...>` and `<Select ...>` tags
* `Def: variable=value` (typically used in combination with the above to initalize a template variable)

I did some refactoring in order to use the same set of _insertion tags_ for text templates as for the HTML forms.

Resolves #375.

I used this as a reference: http://www.philsherrod.com/Winlink/RMS_Express_Message_Templates.pdf.

### DIFF
--- a/internal/forms/forms.go
+++ b/internal/forms/forms.go
@@ -852,15 +852,7 @@ func (m *Manager) fillFormTemplate(tmplPath string, formDestURL string, placehol
 		log.Printf("Warning: unsupported string encoding in template %s, expected utf-8", tmplPath)
 	}
 
-	now := time.Now()
-	validPos := "NO"
-	nowPos, err := m.gpsPos()
-	if err != nil {
-		debug.Printf("GPSd error: %v", err)
-	} else {
-		validPos = "YES"
-		debug.Printf("GPSd position: %s", gpsFmt(signedDecimal, nowPos))
-	}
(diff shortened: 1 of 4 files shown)


internal/forms/forms.go
```diff
@@ -1016,65 +995,170 @@ func (b formMessageBuilder) scanTmplBuildMessage(tmplPath string) (MessageForm,
 	}
 	defer infile.Close()
 
-	placeholderRegEx := regexp.MustCompile(`<[vV][aA][rR]\s+(\w+)\s*>`)
+	placeholderRegEx := regexp.MustCompile(`(?i)<Var\s+(\w+)\s*>`)
```
martinhpedersen, line 998: Unrelated cleanup here. `(?i)` makes the regular expression case-insensitive 🙂

xylo04 (approved):

martinhpedersen:
Thanks for the reviews 😊

PR #445, opened 2024-03-27
Administer Winlink password recovery email

### LINKED ISSUE #442: Add Winlink recovery email address to the config file
To assist new users who loose their password, please add the Winlink recovery email address to the config file when setting up a new account.  The Pat key is configured to allow access to the necessary API. The API can be found at [api.winlink.org](http://api.winlink.org/). A search for 'recovery' will list out the available API and param requirements:
AccountPasswordRecoveryEmailGet	JSON	XML	JSV
AccountPasswordRecoveryEmailSet	JSON	XML	JSV

-Scott, NS7C

martinhpedersen:
* Add CLI command to get/set
* Print notice if not set

Resolves #442


internal/cmsapi/password_recovery.go
```diff
@@ -0,0 +1,95 @@
...
+	return &r
+}
+
+func (r *responseStatus) Error() string { return r.Message }
+
+func PasswordRecoveryEmailGet(ctx context.Context, callsign, password string) (string, error) {
+	req := newJSONRequest("GET", PathAccountPasswordRecoveryEmailGet,
+		url.Values{"callsign": []string{callsign}, "password": []string{password}},
+		nil).WithContext(ctx)
+	resp, err := http.DefaultClient.Do(req)
```
xylo04, line 36: Might be worth setting a custom User Agent when sending this. I think we do so in other places...

xylo04 (approved):

martinhpedersen:
Thanks for the review :) Yes, I think it would be worth setting a custom user agent 😊

PR #467, opened 2025-01-02
Download go modules directly and fix checksum for ptc-go library

perja12:
The checksum for ptc-go library v2.2.3 is not correct for the github version. Also see https://github.com/harenber/ptc-go/issues/38


martinhpedersen:
Hi Per Thomas 😄 

Thanks for reporting this, and for the PR. I wasn't aware of this mismatch... I guess the go mod proxy did it's job and kept builds from breaking 😊  

However, instead of disabling the proxy in our `make.bash` and switching to the other checksum, I'd rather kindly request @harenber to fix this upstream by re-tagging current HEAD of ptc-go as `v2.2.4`. Otherwise builds will fail by default, i.e. if some installs Pat using `go install github.com/la5nta/pat@latest`.

perja12:
Sounds like a good plan. Will close this then 😊

martinhpedersen:
Thank you both 😄

PR #480, opened 2025-05-25
Include and retain X-P2P header for P2P Only messages

elkinsd:
Certain Winlink clients, specifically Winlink Express, expect P2P messages to have the X-P2P header present and set to True, for listing and reporting of P2P messages to work properly. Currently P2P messages received from Pat are not marked as P2P nor reported to the CMS (for US licensed stations).

This is a simple update to prove the concept, and alternatively the change could be to normalize the use of a single X-P2P header for the P2P Only cli and http/ui reporting. Messages received from WLE include the X-P2P header, and could be used to mark as P2P Only in the inbox, etc.

### DIFF
--- a/cli_composer.go
+++ b/cli_composer.go
@@ -77,6 +77,7 @@ func composeMessageHeader(inReplyToMsg *fbb.Message, replyAll bool) *fbb.Message
 		ans := readLine()
 		if strings.EqualFold("y", ans) {
 			msg.Header.Set("X-P2POnly", "true")
+			msg.Header.Set("X-P2P", "true")
 		}
 	case 0:
 		fmt.Println("Message must have at least one recipient")
@@ -191,6 +192,7 @@ func noninteractiveComposeMessage(from string, subject string, attachments []str
 	msg.SetBody(string(body))
 	if p2pOnly {
 		msg.Header.Set("X-P2POnly", "true")
+		msg.Header.Set("X-P2P", "true")
 	}
 
 	postMessage(msg)
--- a/http.go
+++ b/http.go
@@ -315,6 +315,7 @@ func postOutboundMessageHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	if v := r.Form["p2ponly"]; len(v) == 1 && v[0] != "" {
 		msg.Header.Set("X-P2POnly", "true")
+		msg.Header.Set("X-P2P", "true")
 	}
 	if v := r.Form["date"]; len(v) == 1 {
 		t, err := time.Parse(time.RFC3339, v[0])


martinhpedersen:
Hi again,

> Re: option 3, this request was made, but the sole maintainer of WLE will only look at issues that can be recreated solely using WLE to send and receive, and so this will not change.

I'm sorry to hear that 😞 

> Option 2 would be interesting, if the remote station can be detected as "WLE in P2P session" by using the system identifier; if the SID starts with "RMS Express-" it's WLE in P2P (as opposed to "WL2K-" which is an RF/gateway or CMS/telnet connection).

I'm not so sure it's a good idea to special case RMS Express like that. I'd rather we just treat all messages sent over P2P the same way to not complicate things further.

https://github.com/la5nta/wl2k-go/blob/4beed4194d6b4c3ab4bbd35ddf3381e0faeaf637/mailbox/syncdir.go#L158-L165 is where we would need to make changes. If `len(fws) != 0` it means we're in a P2P session, and should set the header.

Somthing like this maybe:

```go
if isP2P := len(fws) != 0; isP2P {
  m.Header.Set("X-P2P", "true")
}
```

> I am not familiar with how messages are stored in-between transfers, to be used with the FBB offset. If the header is added on-the-fly, and an offset given by "WLE in P2P" would be specific to just that remote station and not CMS in general, would the contents of the message to be transferred change between attempts?

Pat stores all messages in the wire format (`b2f`), except for compression. The offset transfer implementation assumes that the messages are immutable. It just loads the file from disk, skips the offset bytes and transmits the rest.

My concern is this:
1. Message 123 is partially transferred via P2P (with the X-P2P header). 200/400 bytes.
2. Message 123 is later sent via CMS (without the X-P2P header). 388/388 bytes. Size differs due to the missing header.
3. The recipient connects to CMS and is notified that Message 123 is pending. Recipient accepts from offset 200, since they have already received the first 200 bytes.
4. The recipient receives the remaining 188 bytes, ending up with a total of 388 *with* the P2P header. 12 bytes are lost, so the message is corrupt and can't be decompressed.

A couple of ideas (brainstorming):
1. Persist the X-P2P header to disk and:
  a) Prevent the message from being sent to CMS later.
  b) Send to CMS regardless. The P2P header would be misleading, but there is no longer any risk of corruption.
2. Generate a new [deterministic) ID for the message when setting the header. This ID would only be visible in P2P sessions.
3. Admit defeat and let Winlink Express deal with the fact that not all P2P messages are correctly flagged in their end 🙈 

I wonder how CMS would react to a `X-P2P: true` header 🤔

elkinsd:
Thanks for your time Martin. This seems like it will just introduce unneeded complexity. Without testing, we can't say how the CMS will react to the header being present (or guarantee it would keep doing the same as it does now), or how WLE would react by seeing a message from the CMS already partially transferred by P2P. It's probably better to just keep the status quo as is.

martinhpedersen:
Thanks for the proposed changes @elkinsd. Ref #487 it seems this not only is a problem flagging the message correctly in Winlink Express' message inbox, but also with their Reply action. Let's continue the discussion in #487.

Thanks again!

Issue #51, opened 2016-08-02
Winlink hybrid: Update MPS list


martinhpedersen:
For users to fully take advantage of the Winlink Hybrid Network, they would need a way to update their MPS (Message Pickup Station) lists.

Winlink Express updates the MPS list via the Winlink Web API directly over HTTP, or by posting a special INQUIRY message containing a sequence of these API requests.

These updates (requests) are guarded by a ```WebServiceAccessCode``` which we need to obtain/generate in order for users to update their MPS list via Pat.

Example update message:
>```
>Message ID: U2YDOUIYDUA4
>Date: 2016/03/27 14:52
>From: LA4TTA
>To: INQUIRY 
>Source: LA4TTA
>Subject: CMSWEBSVC
>
>/mps/delete?requester=LA4TTA&callsign=LA4TTA&WebServiceAccessCode=F630KKDH9L
>/mps/add?requester=LA4TTA&callsign=LA4TTA&mpsCallsign=LA3F&WebServiceAccessCode=F630KKDH9L
>/mps/add?requester=LA4TTA&callsign=LA4TTA&mpsCallsign=SM5RVH&WebServiceAccessCode=F630KKDH9L
>/mps/add?requester=LA4TTA&callsign=LA4TTA&mpsCallsign=HB9AW&WebServiceAccessCode=F630KKDH9L
>/mps/delete?requester=LE1OG&callsign=LE1OG&WebServiceAccessCode=F630KKDH9L
>/mps/add?requester=LE1OG&callsign=LE1OG&mpsCallsign=LA3F&WebServiceAccessCode=F630KKDH9L
>/mps/add?requester=LE1OG&callsign=LE1OG&mpsCallsign=SM5RVH&WebServiceAccessCode=F630KKDH9L
>/mps/add?requester=LE1OG&callsign=LE1OG&mpsCallsign=HB9AW&WebServiceAccessCode=F630KKDH9L
>```
>

*The original WebServiceAccessCode have been changed, in case it contains sensitive information. Thanks to LA4TTA for providing this example.*

Until this is implemented users are advised to update their MPS list using Winlink Express.

martinhpedersen:
Our new API key is authorized for the following APIs:
* MessagePickupStationAdd
* MessagePickupStationDelete
* MessagePickupStationGet
* MessagePickupStationList

Ref issue #110

AquaeAtrae:
[...]

PR #485, opened 2025-06-12
Add mps command for managing message pickup stations

jboverfelt:
ref #51

This implements mps updating (add, delete, list, and list all) using the Winlink API. It takes explicit style inspiration from the account management functions already implemented. It doesn't appear that Winlink actually enforces this, but the API docs say there is a 30 minute rate limit on listing all MPSes so I honored that here. I managed to update and list my own MPS stations successfully over the last few days when testing this.

I did not attempt to implement using an INQUIRY message as described in #51 as I wasn't sure how you would want that to be handled. Perhaps we might use an "--as-message" flag in addition to one of the other options to automatically post to the user's outbox. I do think it would be good to implement as it would allow MPS management without an internet connection.

I also did not attempt to integrate this into the web ui as I wanted to make sure you were happy with the command line implementation first.

Justin AB3E

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,6 +26,7 @@ We welcome your contributions.
 To make the process as seamless as possible, we ask for the following:
 
   - Go ahead and fork the project and make your changes. We encourage pull requests to discuss code changes.
+  - Base your changes off the "develop" branch, not master
   - When you’re ready to create a pull request, be sure to:
       - Run `go fmt`
       - Consider squashing your commits into a single commit. `git rebase -i`. It's okay to force update your pull request.
--- a/internal/cmsapi/mps.go
+++ b/internal/cmsapi/mps.go
@@ -1,0 +1,123 @@
+package cmsapi
+
+import (
+	"context"
+	"encoding/json"
+	"net/url"
+	"regexp"
+	"strconv"
+	"time"
+)
+
+const (
+	PathMPSAdd    = "/mps/add"
+	PathMPSDelete = "/mps/delete"
+	PathMPSGet    = "/mps/get"
+	PathMPSList   = "/mps/list"
+)
+
+// MessagePickupStationRecord represents an MPS record
+type MessagePickupStationRecord struct {
+	Callsign    string     `json:"callsign"`
+	MpsCallsign string     `json:"mpsCallsign"`
+	Timestamp   DotNetTime `json:"timestamp"`
+}
+
+// DotNetTime handles .NET-style JSON date serialization
+type DotNetTime struct{ time.Time }
+
+// UnmarshalJSON implements custom JSON unmarshaling for .NET date format
+func (t *DotNetTime) UnmarshalJSON(b []byte) error {
+	var str string
+	if err := json.Unmarshal(b, &str); err != nil {
+		return err
+	}
+
+	// Handle .NET date format: \/Date(milliseconds)\/
+	re := regexp.MustCompile(`\/Date\((-?\d+)\)\/`)
+	matches := re.FindStringSubmatch(str)
+	if len(matches) == 2 {
+		millis, err := strconv.ParseInt(matches[1], 10, 64)
+		if err != nil {
+			return err
+		}
+		t.Time = time.Unix(millis/1000, (millis%1000)*1000000)
+		return nil
+	}
+
+	// Fall back to RFC3339 format
+	parsedTime, err := time.Parse(time.RFC3339, str)
+	if err == nil {
+		t.Time = parsedTime
+		return nil
+	}
+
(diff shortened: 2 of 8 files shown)


martinhpedersen:
Hi Justin,

Thanks for submitting this PR 😄 I appreciate your decision to focus on the CLI initially, and only via their web API.

I haven't done a full review yet, but it looks very promising!

I'd like to discuss the CLI command/option design. In the newly added `templates` and `account` commands I have introduced subcommand using the `func shiftArgs`. I'm wondering if that would be a better UX for this, in combination with a few command options. I believe it also simplifies the implementation, as we no longer need to count and check for number of options == 1.

```
pat mps list [--all] (defaults to listing mine)
pat mps add W1AW
pat mps delete
```

What do you think? Would it be more intuitive? 🤷 

I'm also wondering if we should rename `delete` to either `clear` or `delete-all` to better communicate that this clears the entire list, and can't be used to delete a single entry. We should probably also print the current list before deleting it, so the user can rebuild their list manually if they accidentally deletes it.

As for the rate limit, I think we need to cache the list so the user can retrieve a copy of the list again without waiting the 30 minutes. But let's focus on the command interface first 😊

jboverfelt:
Didn't mean to push - still working

jboverfelt:
Ok it's ready for review. I realized something in making these changes - "list all" is really a list of everyone's message pickup station configurations, not a list of available MPS. So I changed the behavior to output just the MPS callsigns and a count of how many people use it just like Winlink Express does. 

Changes:

- Rebased on develop, and updated CONTRIBUTING.md to make it clear that's what is desired
- Updated command options
- Changed the list --all behavior to output just MPS callsigns and user counts
- Changed delete to clear to make it obvious that it clears all stations
- Print out previous stations after a clear operation

I didn't implement the caching as discussed. I also felt the test wasn't really providing a ton of value so I removed it. I can add something back in though if you'd like.

mps.go
```diff
@@ -0,0 +1,171 @@
...
+
+func mpsHandle(ctx context.Context, args []string) {
+	mycall := fOptions.MyCall
+	if mycall == "" {
+		fmt.Println("ERROR: MyCall not configured")
+		os.Exit(1)
+	}
+
+	switch cmd, args := shiftArgs(args); cmd {
+	case "list":
```
martinhpedersen, line 37: I think we should keep a local copy of the current list, so the user can review it when they are offline. I imagine there will be a lot of users that don't keep note of their pickup stations.

We could for instance attempt to fetch from the API, write to disk if successful, and fallback to reading from disk. 

What do you think?

CONTRIBUTING.md
```diff
@@ -26,6 +26,7 @@ We welcome your contributions.
 To make the process as seamless as possible, we ask for the following:
 
   - Go ahead and fork the project and make your changes. We encourage pull requests to discuss code changes.
+  - Base your changes off the "develop" branch, not master
```
martinhpedersen, line 29: Thank you! And apologies for not mentioning this sooner 😊

internal/cmsapi/mps.go
```diff
@@ -0,0 +1,123 @@
...
+	"regexp"
+	"strconv"
+	"time"
+)
+
+const (
+	PathMPSAdd    = "/mps/add"
+	PathMPSDelete = "/mps/delete"
+	PathMPSGet    = "/mps/get"
+	PathMPSList   = "/mps/list"
```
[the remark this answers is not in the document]
jboverfelt: Yes please reach out to them when you get a chance. That endpoint does look better (and has no listed rate limit), but it's hard to say for sure since the Pat key isn't authorized for that endpoint. 

It looks like Express does what we're doing currently but it's impossible to say for sure.

mps.go
martinhpedersen: Thats a good point. It's better to direct users to use INQUIRY when we have that in place 👍

martinhpedersen:
@jboverfelt - Haha, yes I noticed. I need to get a U.S license to beat you 😀 

That's perfect. Thank you so much 😊

jboverfelt:
Based on Phil and Lee's email I think we should go with the hybrid station endpoint

martinhpedersen:
Yes, I agree. 

I have merged my refactor branch now, so if you want to continue working on this then please go ahead 😊 I think we can just continue with the open #51 issue.

Highly appreciate your contribution!

martinhpedersen:
You'll find that I have moved MPSHandle and related functions into `cli/mps.go`. The new project structure is not perfect, but I think it's better no structure as it was before ( 🙈 ). I've split the project into three main packages: `cli`, `api` and `app`. The idea is to have core functionality in `app` to be re-used between the cli and api.

PR #488, opened 2025-06-15
Further MPS Improvements

jboverfelt:
* Update MPS list --all to use /hybridStation/list endpoint
* Prevent adding > 3 MPS, warn if adding a third

ref #51

I had a thought regarding the caching of the list --all endpoint. We could avoid a "force-download" flag like rmslist if we just always fetch the latest if we're online (there's no ratelimit). Then, we'd save a copy locally each time. If we're offline, we just reference the latest saved copy and show a warning to the user that it might be stale (and direct them to use an INQUIRY message). I could also be convinced to not cache anything and just direct to use an INQUIRY message on any mps-related failure.

As for the INQUIRY messages, I'm not sure of the best approach there. Should we introduce an "--offline" flag to all the MPS commands that generates an email? Do we want to automatically post that email to the outbox? Do we have good facilities for autogenerating emails already?

### DIFF
--- a/cli/mps.go
+++ b/cli/mps.go
@@ -2,11 +2,9 @@ package cli
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"strings"
-	"time"
 
 	"github.com/la5nta/pat/app"
 	"github.com/la5nta/pat/internal/cmsapi"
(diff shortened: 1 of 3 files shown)


martinhpedersen:
> I had a thought regarding the caching of the list --all endpoint. We could avoid a "force-download" flag like rmslist if we just always fetch the latest if we're online (there's no ratelimit). Then, we'd save a copy locally each time. If we're offline, we just reference the latest saved copy and show a warning to the user that it might be stale (and direct them to use an INQUIRY message). I could also be convinced to not cache anything and just direct to use an INQUIRY message on any mps-related failure.

I think that's good idea! My only reservation is, at least for me, I often execute the command only to realize I need to filter it using grep to find what I'm looking for. So having a cache with short TTL of a minute or two would be nice IMHO. But we can always add this later, maybe after implementing the web gui part so we know the requirements there.

-----

> As for the INQUIRY messages, I'm not sure of the best approach there. Should we introduce an "--offline" flag to all the MPS commands that generates an email? Do we want to automatically post that email to the outbox? Do we have good facilities for autogenerating emails already?

Yes, I think we need a flag. I'd prefer a more explicit name though, something like `--via-radio` or `--message-inquiry` maybe? I also wonder if it would be best to add this to the mps part of the command. `pat mps --message-inquiry [add/list/clear]`? 🤔  I do think it would be appropriate to automatically post the email to the outbox.

The `position` command is an example of a command that generates a message similar to this.

We currently don't have anything that parses a received message to extract the requested data. That will be an interesting feature to add. I've been playing around with the idea to automatically extract the generated password when a new user connects for the first time, to automatically store it in the config and also prompt the user to write it down as a backup. We could use the same mechanism for that maybe 😊

martinhpedersen:
Btw, I noticed that the two booleans appear to be mutually exclusive. It makes sense that a MPS is either manually *or* automatically forwarding messages. I'm wondering if we should merge these fields into a single field in the UI? And maybe also add the option to filter based on this. `--automatic-only` or something. What do you think?

Adding to this, I think we should also consider enriching the rms list with this data at some time or visa versa. This is the 'H' column in Winlink Express. I believe it makes sense to factor in the available band and mode when deciding which MPS to use.

PR #512, opened 2025-12-22
Portability patch

TheMetallists:
This is a proposal to add "portability library" build mode to PAT, which paves the way for Android support.
A few notes:

- To support multiple entry points, i had to move the web/ folder into a subpackage, so different "main" packages could embed the same gui.
- Inside the web/ folder i found some issue with include paths inside html files - they were relative and didn't include the /dist directory through which all the static files were exposed by the http server. This might be a misunderstanding by me of some underlying logic, so feel free to correct me on this.
- Also, i added a launcher (under golang's seemingly-standardized folder "cmd") for the library, so people without Android tools/experience could debug the libraray's entry point. I am not insisting on it being necessary.

So, let me know what you think, what should be improved or changed.


libportapat/config_related.go
```diff
@@ -0,0 +1,66 @@
...
+	"math/rand"
+	"os"
+	"strings"
+)
+
+func SetRandomizePort(rndport bool) {
+	flagRandomisePort = rndport
+}
+
+func doRandomisePort() error {
```
martinhpedersen, line 19: Instead of setting the custom port by rewriting the config file, you have two other options:
1. Pass `-addr [host:port]` as arguments to the `http` command.
2. Set env var `PAT_HTTPADDR=[host:port]` (this applies to all config options)

By doing this, you don't need to read/write the config file.

libportapat/porta_paths/paths.go
```diff
@@ -0,0 +1,86 @@
+package porta_paths
```
martinhpedersen, line 1: Setting custom paths can be done with standard XDG environment variables (`XDG_DATA_HOME`, `XDG_CONFIG_HOME` and `XDG_STATE_HOME`).

This should eliminate the need for the `porta_paths` package I believe?

martinhpedersen (commented):
Thanks for the PR 😊 

To be honest, I'm a bit hesitant to merge this as it is right now, mainly because the code is very specific to your project and it's branding. It's seems like bad practice to me, especially with regards to Separation of Concerns.

That said, I believe we can fulfill your specific requirements while still maintaining good SoC between the two projects - we just need to make some small adjustments to a few packages in Pat.

Please see the inline comments regarding setting custom HTTP listen addr and data dirs.

The way I see it, you have 4 main requirements:
- [x] Set custom HTTP listen addr
- [x] Set custom config/state/log dirs
- [x] Get the version string
- [x] Load the embedded fs

I believe the last two is the only remaining requirements that currently requires you to access internal packages/files. It is already possible to `import github.com/la5nta/pat/app` and other non-internal packages.

I think we can expose the version information through a proxy method in app package. And as for the embedded fs: I think that can be achieved by having a `web/embedded_fs.go` file and let the api package import that new `package web` directly.

What do you think about this? Is there anything else that I might have missed?

Thanks 😊

TheMetallists:
This is just a suggestion, feel free to change anything you like.
Regarding environment variables and os.Args - i'll give it a try.
For the port it is indeed a better solution, for the paths - well if it works, it is also futureproof.

Regarding the version string - this function can be removed, i forgot to.

How do you see bringing Android support into the mainline? Is it very low priority and should stay as a separate project for now, or should i remove all the "different project" branding?

martinhpedersen:
> How do you see bringing Android support into the mainline? Is it very low priority and should stay as a separate project for now, or should i remove all the "different project" branding?

I'm not sure... I think maybe having it as a separate repo would make sense, or possibly a `android/` dir in this repo if the complexity is low. We would also need to add a step to build the apk in https://github.com/la5nta/pat-build to keep the release process effective.

I don't think we should make it the highest priority right now, but I'm motivated to spend some time on it :)

If it's ok with you, maybe we can continue with portapat (and libportapat) as an experimental side project with the goal to eventually move it into the main repo or as a side repo under the same organization?

If so, I need to kindly request that you relicense it under MIT. We can't merge code that is copied from a GPL licensed repo, and relicensing down the road might be a challenge if you get contributions from other people aswell.

martinhpedersen:
With commit 946a91fcc4765d8fa8107448008f037ae2e16f64 I believe all four requirements are met, so we should now support having a android specific main package outside this repo that is able to start the web gui.

I have also fixed the missing "Back to mailbox" button in the Configuration page on small screens. Haven't rebuilt the web assets yet, but will do so before the next release 😊

TheMetallists:
Ok, as time permits I'll relicense and megre/change/rerequest accordingly.

martinhpedersen:
Thank you! Looking forward to it 😊 Please let me know how the recent changes on `develop` works for you, when you find some time. Thanks again!

Issue #537, opened 2026-08-07
FR: Integrate ARDOP WEBGUI


nathanielbutts:
ARDOP has an option from the command line to launch a webgui on http://localhost:8514 which is very useful for troubleshooting, adjustment, etc.  I have a script to start ardop and pat like ```ardopcf -G 8514```, or ```ardopcf --webgui 8514```.  

My favorite option  would be to have an option to launch and display this webgui output  (waterfall, phasor diagram, adjustments) while attempting to connect.  

My second favorite option would be to launch this from Pat so it doesn't need to be done separately, and then display a link or open that tab in the browser.

https://github.com/pflarue/ardop/blob/master/docs/USAGE_linux.md

mwthomson:
I have been thinking something similar, @nathanielbutts. I start Pat and then have to start each transport separately. 

I don't love the idea of trying to incorporate the ARDOP UI into Pat, though. If the Ardop guys make a change there's risk it would break Pat, or not display properly. But, I do love the idea of adding launchers per transport. This would require the user to input the command to launch the transport in the transport config. IF the value is present, Pat shells to the command line to launch the transport.

Does that sound right?

mwthomson:
@nathanielbutts - Would you mind looking at #544 and providing your feedback?

PR #544, opened 2026-08-25
Add optional launch_cmd to auto-start Ardop/VARA/AGWPE TNC daemons

mwthomson:
Addresses #537.

Adds an optional `launch_cmd` setting to Pat's config for the four transports that depend on an external TNC daemon: Ardop, VaraHF, VaraFM, and AX.25/AGWPE. When set, Pat spawns that daemon itself the first time the transport is used, instead of requiring the user to start it by hand first.

```json
"ardop": {
  "addr": "localhost:8515",
  "launch_cmd": {"path": "ardopcf", "args": ["--webgui", "8514"]}
}
```

Behavior, consistent across all four transports:
- Skips the launch if the daemon is already reachable (no duplicate spawns, even under concurrent connect/listen).
- On outgoing connect, waits up to 5s (polling every 500ms) for a freshly-launched daemon to come up before giving up.
- On incoming listen, relies on the existing `ListenerHub` retry loop rather than blocking.
- Relaunches automatically if the daemon crashes or is killed mid-session.
- Terminates any process it spawned when Pat itself exits (`App.Close()`) — no orphaned daemons.
- Logs a terse `Pat failed to launch <transport>` on failure; full error only at debug level.
- Strictly opt-in — leaving `launch_cmd` unset is a complete no-op, existing behavior unchanged.

This directly solves the underlying need in #537 (bringing up `ardopcf`'s webgui alongside Pat) without Pat embedding or knowing anything about that UI — it just launches whatever command is configured.

## Test plan
- [x] `go build ./...`, `go vet ./...` clean
- [x] `go test ./...` passes
- [x] Unit tests cover: skip-when-reachable, launch-then-connect within poll budget, launch failure (bad path / never reachable) with exact log message, no duplicate spawn under concurrent Connect+Listen (`-race`), relaunch after crash, `App.Close()` terminates spawned processes
- [x] Manually tested against a running AGWPE/Direwolf setup


mwthomson:
While working on a follow-up transport driver I found two config-validation gaps in this PR and pushed a fix:

- #548 — Ardop never got the default-Addr rescue that AGWPE/VaraHF/VaraFM have, so a config setting only `ardop.launch_cmd` (no `addr`) launches `ardopcf` successfully but then fails every dial.
- #549 — `VaraConfig.UnmarshalJSON`'s port-validation guard had narrowed to just `Addr != ""`, silently accepting a config with e.g. `rig`/`ptt_ctrl` set but no `addr` instead of failing fast at load.

Both fixed in the latest commit.

mwthomson:
General cleanup and fixes for #547, #548, and #549 now rolled into this PR. Ready for review

Issue #534, opened 2026-07-27
Correct frequency capitalization from "KHz" to "kHz" in UI menus


t-o-k:
Please update the user interface menus to use lowercase kHz instead of uppercase KHz.

According to the International Bureau of Weights and Measures (BIPM) SI Brochure (and ISO 80000-3), the SI prefix for kilo is a lowercase k.

"The International System of Units - 9th edition (2019)", page 138
https://www.bipm.org/en/si-brochure-9

The BIPM brochure states: "With the exception of da (deka), h (hecto) and k (kilo), all multiple prefix symbols are upper-case letters".

Therefore, kHz must be lowercase, while MHz, GHz, and THz must be uppercase.

kHz = 1000 Hz

An uppercase K represents kelvin (thermodynamic temperature), making "KHz" mathematically incorrect (Kelvin-Hertz).

Also see: "NIST Special Publication 330 - The International System of Units (SI)", page 23
https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.330-2019.pdf

t-o-k:
I did a quick scan of the codebase and found that "KHz" is used in several places across the Go backend and UI templates. 

Here are the occurrences that can be adjusted to align with the correct `kHz` standard (or `Khz` where Go export conventions require a capitalized first letter):

* **UI Display:** `web/src/index.html` (and the built `web/dist/index.html`)
* **Go Backend:** `app/freq.go` and `app/rmslist.go`

**Grep output for reference:**
```text
./app/freq.go:          KHz  json.Number `json:"khz"`
./app/freq.go:          KHz:  json.Number(fmt.Sprint(f.KHz())),
./app/freq.go:func (f Frequency) KHz() float64 { return float64(f) / 1e3 }
./app/rmslist.go:       chURL, _ := url.Parse(fmt.Sprintf("%s:///%s?freq=%v", toTransport(gc), targetCall, freq.KHz()))
./web/src/index.html:                          <div class="input-group-addon">KHz</div>
./web/dist/index.html:                          <div class="input-group-addon">KHz</div>
```

*(Note: For the Go fields and methods, switching from `KHz` to `Khz` fixes the lowercase "h" while keeping the code exported. Depending on whether strict SI casing is wanted in the JSON payloads or strict lowercase API conventions, the JSON tag could also be updated from `json:"khz"` to `json:"kHz"`).*

martinhpedersen:
Takk Tor Olav 😄

PR #551, opened 2026-08-31
Correct frequency unit label to SI-standard "kHz"

mwthomson:
Fixes #534.

The connect dialog's frequency input labelled its unit `KHz`. Corrected the capitalization to `kHz` for kilohertz.

This is the only miscased frequency unit in the web UI — `Frequency.String()` already emits `MHz`, and nothing else renders one.

`web/dist/` left for a separate asset rebuild, per project convention.


martinhpedersen:
Thank you 😀

